PR #2, opened 2022-11-09
Colin's prototype

semicolin:
Copied the code from my prototype


projects/showcase/src/app/datasets/example.service.ts
```diff
@@ -0,0 +1,51 @@
...
+          type: 'time',
+          time: {
+            unit: 'day',
+            displayFormats: {
+              day: 'd MMM yyyy',
+            },
+          },
+        },
+      },
+    }).pipe(delay(2000));
```
yashpmohod, line 50: why do we need a delay here?

semicolin: This is example service is just for testing. I was trying to simulate the delay that you would get when making a network request in a real service.

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-manager.service.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  // todo: roll this into stateSubject?
+  private timelineRangeSubject = new ReplaySubject<{ min: number; max: number }>();
+  timelineRange$ = this.timelineRangeSubject.asObservable();
+
+  loading = false;
+
+  retrieveAll() {
+    this.loading = true;
+    merge(...this.dataLayerServices.map((service) => service.retrieve())).subscribe({
```
yashpmohod, line 58: @semicolin, can you please add comment around this, how it exactly performs

semicolin: added a comment

yashpmohod (approved):
as such, there is no comment from my side,  going to approve, if I want I will ask to clin

PR #4, opened 2022-11-29
Build and Test Workflow

semicolin:
Adds workflow to build & test the library and showcase app


semicolin:
Workflow is failing because bundle size exceeds Angular's default budget. I will investigate.

semicolin:
I don't see anything wrong with the bundle. I guess the default budget is just too small for an Angular Material app without lazy-loaded modules. I will increase the budget for now until we implement lazy-loading.

hankwallace (approved):

PR #8, opened 2022-12-06
RT-508:Annotation options

amolht:
## Overview

-   Added setting for annotation options 

  (show/hide, change color, change label, change yMax, YMin)

## How it was tested

- Tested on local by running a app
- Ran unit test locally.

## Checklist


amolht:
Screenshot from 2022-12-06 20-40-10

projects/ngx-charts-on-fhir/src/lib/data-layer-list/dataset-annotatitons/dataset-annotations.component.ts
```diff
@@ -0,0 +1,65 @@
...
+import produce from 'immer';
+import { merge } from 'lodash-es';
+import { DataLayerColorService } from '../../data-layer/data-layer-color.service';
+
+@Component({
+  selector: 'dataset-annotations',
+  templateUrl: './dataset-annotations.component.html',
+  styleUrls: ['./dataset-annotations.component.css'],
+})
+export class DatasetAnnotationsComponent implements OnInit {
```
semicolin, line 12: Please rename this component to `AnnotationOptionsComponent` (including selector and filenames)

projects/ngx-charts-on-fhir/src/lib/data-layer-list/dataset-options/dataset-options.component.css
```diff
@@ -13,3 +13,7 @@ mat-form-field,
 mat-slide-toggle {
   display: block;
 }
+
+.example-full-width {
+  width: 50%;
+}
```
semicolin, line 19: unused?

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.ts
```diff
@@ -22,6 +22,37 @@ export class DataLayerColorService {
...
+  chooseAnnotationColorsFromPalette(layer: DataLayer): void {
+    if(layer.annotations){
+      for (let annotation of layer.annotations) {
+        if (!this.getAnnotationColor(annotation)) {
+          const color = this.palette[this.nextColorIndex];
+          this.nextColorIndex = (this.nextColorIndex + 1) % this.palette.length;
+          this.setAnnotationColor(annotation, color);
+        }
+      }
+    }
```
semicolin, line 34: For now, I think I would prefer to leave the default light blue color for all annotations unless user chooses a different one. At least until we have a way to link annotation to a dataset and use matching colors.

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.ts
```diff
@@ -22,6 +22,37 @@ export class DataLayerColorService {
...
+   
+  }
+  setAnnotationColor(annotation: any, color: string): void {
+    const line = annotation;
+    // line.borderColor = color;
+    line.backgroundColor = color + '33'; // temporary dirty hack to set opacity. assumes color is in 6-character hex format.
+    // line.pointBorderColor = color;
+    // line.pointBackgroundColor = color;
+  }
```
semicolin, line 43: please remove commented-out code

projects/ngx-charts-on-fhir/src/lib/fhir-mappers/simple-observation-mapper.service.ts
```diff
@@ -41,7 +41,7 @@ export class SimpleObservationMapper implements Mapper<SimpleObservation> {
   ) {}
   canMap = isSimpleObservation;
   map(resource: SimpleObservation): DataLayer {
-    return {
+    const output =  {
```
semicolin, line 44: why assign this to a variable before returning it?

projects/ngx-charts-on-fhir/src/lib/fhir-mappers/simple-observation-mapper.service.ts
```diff
@@ -64,12 +64,14 @@ export class SimpleObservationMapper implements Mapper<SimpleObservation> {
       },
       annotations: resource.referenceRange?.map<ChartAnnotation>((range) =>
         merge({}, this.annotationOptions, {
+          display: true,
```
semicolin, line 67: should we add `display: true` to the default ANNOTATION_OPTIONS?

semicolin (changes requested):
In addition to my comments below, please fix failing tests and run sonarlint and prettier.

semicolin:
Please revert the change to pull_request_template.md and add it to .prettierignore

projects/ngx-charts-on-fhir/src/lib/fhir-mappers/simple-observation-mapper.service.ts
amolht: no @semicolin we don't need to display: true explicitly, so I have removed it.

projects/ngx-charts-on-fhir/src/lib/data-layer-list/data-layer-list/data-layer-list.component.ts
```diff
@@ -39,7 +39,8 @@ export class DataLayerListComponent {
   }
 
   onLayerChange(layer: ManagedDataLayer) {
-    this.layerManager.update(layer);
+    if(layer.id)
+       this.layerManager.update(layer);
   }
```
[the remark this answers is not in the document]
amolht: @semicolin when we change any label and then click on the color change we get that id falsy

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.spec.ts
```diff
@@ -19,10 +19,10 @@ const mockColorService = {
...
 })
 class MockColorPickerComponent implements ControlValueAccessor {
-  writeValue(obj: any): void {}
-  registerOnChange(fn: any): void {}
-  registerOnTouched(fn: any): void {}
-  setDisabledState?(isDisabled: boolean): void {}
+  writeValue(obj: any): void { /* TODO document why this method 'writeValue' is empty */ }
+  registerOnChange(fn: any): void { /* TODO document why this method 'registerOnChange' is empty */ }
+  registerOnTouched(fn: any): void { /* TODO document why this method 'registerOnTouched' is empty */ }
+  setDisabledState?(isDisabled: boolean): void { /* TODO document why this method 'setDisabledState' is empty */ }
```
semicolin, line 25: If we really need to document why the mock has empty methods, please do so. I don't think we need TODOs for that.

projects/ngx-charts-on-fhir/src/lib/data-layer-list/data-layer-list/data-layer-list.component.ts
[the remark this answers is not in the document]
semicolin: Ok, I checked in the debugger. The layer id does not change. This handler is catching a change event bubbling up from the input element. We should use more specific event names that don't collide with built-in events. You did it correctly in `AnnotationListComponent`. The problem here is with `DataLayerOptionsComponent`. I will open a new issue for fixing that. You can leave this layer.id check here.

projects/ngx-charts-on-fhir/src/lib/fhir-mappers/simple-observation-mapper.service.ts
semicolin: looks like it's still there. do you have more commits to push?

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-list/annotation-list.component.html
```diff
@@ -0,0 +1,22 @@
...
+<div class="scrolling-list">
+  <mat-accordion displayMode="flat" multi="true">
+    <mat-expansion-panel *ngFor="let annotation of _annotations; trackBy: trackByIndex" expanded="true">
+      <mat-expansion-panel-header>
+        <mat-panel-title>
+          <mat-checkbox
+            (click)="$event.stopPropagation()"
+            (change)="onCheckboxChange(annotation, $event)"
+            [checked]="annotation.display"
+            [title]="annotation.display ? 'Show this dataset on the chart' : 'Hide this dataset on the chart'"
```
semicolin, line 12:
```suggestion
            [title]="annotation.display ? 'Show this annotation on the chart' : 'Hide this annotation on the chart'"
```

projects/ngx-charts-on-fhir/src/lib/fhir-mappers/simple-observation-mapper.service.ts
amolht: I have updated PR, removed  `display: true` from simple-observation-mapper.service.ts

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.spec.ts
amolht: sonarcloud show a critical error for an empty method, so added a comments for resolving sonarcloud issue

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.spec.ts
```diff
@@ -19,10 +19,10 @@ const mockColorService = {
...
 })
 class MockColorPickerComponent implements ControlValueAccessor {
-  writeValue(obj: any): void { /* TODO document why this method 'writeValue' is empty */ }
-  registerOnChange(fn: any): void { /* TODO document why this method 'registerOnChange' is empty */ }
-  registerOnTouched(fn: any): void { /* TODO document why this method 'registerOnTouched' is empty */ }
-  setDisabledState?(isDisabled: boolean): void { /* TODO document why this method 'setDisabledState' is empty */ }
+  writeValue(obj: any): void { /* Writes a new value to the element. */ }
+  registerOnChange(fn: any): void { /* Registers a callback function that is called when the control's value changes in the UI. */ }
+  registerOnTouched(fn: any): void { /* Registers a callback function that is called by the forms API on initialization to update the form model on blur. */ }
+  setDisabledState?(isDisabled: boolean): void { /* Function that is called by the forms API when the control status changes to or from 'DISABLED'. Depending on the status, it enables or disables the appropriate DOM element */ }
```
semicolin, line 25: Sonar flags empty methods because they might be a mistake. The purpose of adding comments to empty methods is to let other developers (and sonar) know that it's not a mistake. Your comments look like they were pulled from the interface documentation and are rather confusing in this context. A simple comment like `/* intentionally empty stub */` would make more sense here.

amolht: @semicolin will correct it, overthinking to add comments for it.

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-list/annotation-list.component.spec.ts
```diff
@@ -21,4 +21,87 @@ describe('DatasetAnnotationListComponent', () => {
   it('should create', () => {
     expect(component).toBeTruthy();
   });
+
+  it('when onAnnotationsChange called it should emit event', () => {
```
semicolin, line 25: FYI: The jasmine function is named `it` so you can use it as the first word in the sentence. Sort of like a writing prompt to help us write BDD-style expectations. For example `it('should do X', ...)` reads as "it should do X".

When testing a class, I like to create a `describe()` for each method. For example:
```
describe('AnnotationListComponent', () => {
  describe('onAnnotationsChange', () => {
    it('should emit annotationsChange event', ...);
  });
});
```

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.spec.ts
```diff
@@ -47,4 +55,57 @@ describe('AnnotationOptionsComponent', () => {
   it('should create', () => {
     expect(component).toBeTruthy();
   });
+
+  it('called private updateModel method', () => {
+    let spyon = spyOn<any>(component, 'updateModel');
+    let formValue: any = { color: '#377eb8', label: 'Systolic Blood Pressure Reference Range', yMax: 130, yMin: 90 };
+    component['updateModel'](formValue);
```
semicolin, line 62: 1. Please don't bypass `private` access modifiers in unit tests. Unit tests should be testing the public API of the component.
2. This test never executes any real code. It creates a spy, calls the spy, and verifies that that the spy was called.

semicolin (changes requested):

yashpmohod (dismissed):

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-list/annotation-list.component.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+  describe('onCheckboxChange', () => {
+    it('should update view, control and emit the option if an option is clicked', () => {
+      spyOn(component.annotationsChange, 'emit');
+      const annotation={ "label": { "display": true, "position": { "x": "start", "y": "end" }, "color": "#666666", "font": { "size": 16, "weight": "normal" }, "content": "Systolic Blood Pressure Reference Range" }, "type": "box", "backgroundColor": "#ECF0F9", "borderWidth": 0, "drawTime": "beforeDraw", "display": false, "yScaleID": "mm[Hg]", "yMax": 130, "yMin": 90 };
+      component.annotations = [annotation];
+      let event: any= {
+         checked : true
+      }
+      component.onCheckboxChange(annotation, event);
+      expect(component.annotationsChange.emit).toHaveBeenCalled();
```
semicolin, line 33: I see a number of issues with this test:
1. It is testing a `{ checked: true }` event for an annotation that is already set to `{ display: true }`
2. It is not checking the emitted value. I guess it's worthwhile to check that it's emitting _something_, but it would be better to check that it's emitting the _correct_ thing.
3. It is using a spy to stub the component's own `EventEmitter`. This seems like a bad idea. Why not just subscribe to the real `EventEmitter`?
4. The annotation has a lot of irrelevant properties. Please only include what is needed to make the test pass.
5. It is ignoring the component's template and testing it like a regular class. It's ok to do this sometimes, but it looks like all of your tests follow this pattern. I think we should write some tests that interact with the DOM so we'll know that everything in the template is wired up correctly. Angular Material provides test harnesses to make this easy. 

Here is how I would rewrite this test to solve all of these issues:
```ts
    it('should set display=true when checked', async () => {
      const original = [{ label: { content: 'Test' }, display: false }];
      const expected = [{ label: { content: 'Test' }, display: true }];
      let emitted: any = null;
      component.annotations = original;
      component.annotationsChange.subscribe((e) => (emitted = e));
      const checkbox = await loader.getHarness(MatCheckboxHarness);
      await checkbox.check();
      expect(emitted).toEqual(expected);
    });
```

projects/ngx-charts-on-fhir/src/lib/fhir-mappers/component-observation-mapper.service.ts
```diff
@@ -74,6 +74,7 @@ export class ComponentObservationMapper implements Mapper<ComponentObservation>
         (component) =>
           component.referenceRange?.map((range) =>
             merge({}, this.annotationOptions, {
+              display: true,
```
semicolin, line 77: didn't you say `display: true` is not needed?

semicolin (changes requested):
Sorry that you got stuck writing tests for my code. Maybe SonarCloud just doesn't like you 😆.

This is great progress. Let's keep iterating on these tests until we have a "best practice" pattern that we can reuse for future components. See my comments below. I didn't comment on every test case, but many of my suggestions for the checkbox test can be applied to others as well.

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-list/annotation-list.component.spec.ts
amolht: Hi @colin we have tried this when I am running the test case we are getting emitted as a null and the assertion is failing because expect(emitted).toEqual(expected) run before subscribe. it's not waiting to subscribe to the event.
I have coverted like this and it is working

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    expect(component).toBeTruthy();
+  });
+
+  it('form value should update from form changes', fakeAsync(() => {
+    let annotaionOptions = { color: 'color', label: 'label', yMax: 80, yMin: 120 };
+    updateForm(annotaionOptions);
+
+    fixture.detectChanges();
+    const updateModelSpy = spyOn<any>(component, 'updateModel');
+    component['updateModel'](annotaionOptions);
```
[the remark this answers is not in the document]
amolht: because I was trying to test the update model.

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-list/annotation-list.component.spec.ts
semicolin: In your solution, the checkbox will not get checked until after the annotation changes. What will cause the annotation to change? I suspect your test is passing because `annotationsChange` never emits anything and the test completes without making any assertions.

I don't understand why my test is not working for you. The Angular Material docs say:
> The harness also automatically waits for the fixture to be stable, which will cause the test to wait for setTimeout, Promise, etc.

So it should not get past the `await checkbox.check()` until everything is resolved.

semicolin: We should probably enable jasmine's `failSpecWithNoExpectations` option so tests cannot pass without making an assertion.

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+      const annotaionFormLabel = fixture.nativeElement.querySelector('#annotationform').querySelectorAll('input')[1];
+      annotaionFormLabel.value = 'somevalue';
+      annotaionFormLabel.dispatchEvent(new Event('input'));
+      fixture.detectChanges();
+      fixture.whenStable().then(() => {
+        const label = component.form.get('label');
+        expect(annotaionFormLabel.value).toEqual(label?.value);
+      });
+    });
+  });
```
semicolin, line 81: This call to `fixture.whenStable()` leaves a dangling promise. Tests that use this pattern need to use [waitForAsync](https://angular.io/guide/testing-components-scenarios#async-test-with-waitforasync), but I recommend rewriting it to use [MatInputHarness](https://material.angular.io/components/input/api#MatInputHarness) instead for better readability.

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    await TestBed.configureTestingModule({
+      providers: [
+        { provide: DataLayerColorService, useValue: colorService },
+        { provide: COLOR_PALETTE, useValue: ['#000000', '#ffffff'] },
+      ],
+    }).compileComponents();
+  });
+
+  describe('chooseColorsFromPalette', () => {
+    it('should call chooseColorsFromPalette', inject([DataLayerColorService], (service: DataLayerColorService) => {
```
semicolin, line 19:
```suggestion
    it('should call setColor', inject([DataLayerColorService], (service: DataLayerColorService) => {
```

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+        annotations: [{ label: { display: true } }],
+        backgroundColor: '#ECF0F9',
+      };
+      let setColorSpy = spyOn(service, 'setColor');
+      service.chooseColorsFromPalette(layer);
+      expect(setColorSpy).toHaveBeenCalled();
+    }));
+  });
+  describe('getAnnotationColor', () => {
+    it('should get correct annotation color by calling getAnnotationColor', inject([DataLayerColorService], (service: DataLayerColorService) => {
```
semicolin, line 31:
```suggestion
    it('should return the annotation background color', inject([DataLayerColorService], (service: DataLayerColorService) => {
```

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+        label: 'Diastolic Blood Pressure',
+        yAxisID: 'mm[Hg]',
+        data: [],
+        backgroundColor: '#e41a1c33',
+      };
+      const getColorSpy = spyOn(colorService, 'getColor');
+      colorService.getColor(dataset);
+      expect(getColorSpy).toBeDefined();
+      expect(getColorSpy).toHaveBeenCalled();
+    });
```
semicolin, line 52: @amolht @yashpmohod
This test is not testing anything. It is calling a stub.
Tests should not spyOn/stub/mock the code that they are testing.
Test should not make direct calls to their own spies.

The purpose of spies/stubs/mocks is to replace the code's _dependencies_ with simplified ones that we can observe and control.

Test should call the _real_ code. The real code will call the spied-on dependencies. Then the test can check that the correct dependencies were called.

semicolin (changes requested):

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.spec.ts
```diff
@@ -1,96 +1,76 @@
...
-      expect(component.form.value).toEqual(annotationOptions);
-    });
-  }));
+  it('should update annotation label when input is changed', async () => {
+    let labelInputHarness = await loader.getHarness(MatInputHarness.with({ selector: "[id='annotationLabel']" }));
+    let labelInputControl = fixture.componentInstance.form.controls['label'];
+    const testLabel = 'test label';
+    await labelInputHarness.setValue(testLabel);
+    expect(labelInputControl.value).toBe(testLabel);
+  });
```
semicolin, line 44: This is an improvement because it is now testing that the template is wired up to the reactive form.

But it is still not testing that the model gets updated correctly.

`component.form`, although it is a public property, is not really part of the component's public API. That's just an implementation detail. Instead, the test should check the value emitted by `annotationChange` event.

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.ts
```diff
@@ -0,0 +1,66 @@
...
+export class AnnotationOptionsComponent implements OnInit {
+  private _annotation?: any;
+  newAnnotation?: any;
+
+  @Input() set annotation(annotation: any) {
+    this._annotation = annotation;
+    this.updateForm(annotation);
+  }
+
+  @Output() annotationsChange = new EventEmitter<any>();
```
semicolin, line 21: This is now the exact same event name as annotation-list component and we could run into similar issue with event name collisions. It's also a bit misleading because the output name is plural but it's only emitting a single annotation. Can you please change this to `annotationChange` (singular)?

semicolin (changes requested):

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+  let component: AnnotationOptionsComponent;
+  let fixture: ComponentFixture<AnnotationOptionsComponent>;
+  let colorService: DataLayerColorService;
+  let palette: string[] = ['#FFFFFF', '#121212', '#000000'];
+  let loader: HarnessLoader;
+
+  beforeEach(async () => {
+    colorService = new DataLayerColorService(palette);
+    await TestBed.configureTestingModule({
+      imports: [ReactiveFormsModule, NoopAnimationsModule, MatInputModule, MatFormFieldModule, ColorPickerModule],
```
semicolin, line 23: why did you decide to use the real color picker instead of a mock?

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.spec.ts
semicolin:
```suggestion
  it('should update annotation label when input is changed', async () => {
    let emitted: any = null;
    component.annotation = { label: { content: 'Old' } };
    component.annotationsChange.subscribe(e => emitted = e)
    const labelInputHarness = await loader.getHarness(MatInputHarness.with({ selector: "[id='annotationLabel']" }));
    await labelInputHarness.setValue('New');
    expect(emitted.label.content).toEqual('New');
  });
```

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+    let labelInputHarness = await loader.getHarness(MatInputHarness.with({ selector: "[id='annotationLabel']" }));
+    let yMinInputHarness = await loader.getHarness(MatInputHarness.with({ selector: "[id='yMin']" }));
+    let yMaxInputHarness = await loader.getHarness(MatInputHarness.with({ selector: "[id='yMax']" }));
+    expect(labelInputControl.value).toBe('Test');
+    expect(yMaxInputControl.value).toBe(200);
+    expect(yMinInputControl.value).toBe(100);
+    expect(await labelInputHarness.getValue()).toBe('Test');
+    expect(await yMinInputHarness.getValue()).toBe('100');
+    expect(await yMaxInputHarness.getValue()).toBe('200');
+  });
```
semicolin, line 75: There's no need to check the values of the reactive form controls. 

```suggestion
  it('should update form when input is change', async () => {
    let annotation = { label: { content: 'Test' }, yMin: 100, yMax: 200 };
    component.annotation = annotation;
    let labelInputHarness = await loader.getHarness(MatInputHarness.with({ selector: "[id='annotationLabel']" }));
    let yMinInputHarness = await loader.getHarness(MatInputHarness.with({ selector: "[id='yMin']" }));
    let yMaxInputHarness = await loader.getHarness(MatInputHarness.with({ selector: "[id='yMax']" }));
    expect(await labelInputHarness.getValue()).toBe('Test');
    expect(await yMinInputHarness.getValue()).toBe('100');
    expect(await yMaxInputHarness.getValue()).toBe('200');
  });
```

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-list/annotation-list.component.spec.ts
```diff
@@ -26,37 +26,40 @@ describe('DatasetAnnotationListComponent', () => {
...
-      component.annotations = [annotation];
-      let event: any = { checked: true };
-      component.onCheckboxChange(annotation, event);
-      expect(component.annotationsChange.emit).toHaveBeenCalled();
+    it('should set display true when checked', async () => {
+      let emitted: any;
+      component.annotationsChange.subscribe((e) => (emitted = e));
+      const annotations = [{ label: { content: 'Test', display: true }, display: false }];
+      component.annotations = annotations;
+      let checkedEvent: any = [{ label: { content: 'Test', display: true }, display: true }];
```
semicolin, line 34: this is a misleading variable name. It is not a checkbox 'checked' event, it's the expected value of the annotation.

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-list/annotation-list.component.spec.ts
```diff
@@ -26,37 +26,40 @@ describe('DatasetAnnotationListComponent', () => {
...
+      component.annotations = annotations;
+      let unCheckedEvent: any = [{ label: { content: 'Test', display: true }, display: false }];
+      let checkBoxHarness = await loader.getHarness(MatCheckboxHarness.with({ selector: "[id='annotationDisplayCheckBox']" }));
+      await checkBoxHarness.uncheck();
+      expect(emitted).toEqual(unCheckedEvent);
     });
   });
 
   describe('onAnnotationsChange', () => {
     it('should update view, control and emit the option if an option is clicked', () => {
```
semicolin, line 52: please rewrite this description to reflect the behavior that is being tested. there are no options being clicked in this test.
how about "should emit annotationsChange event with updated annotations"

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-list/annotation-list.component.html
```diff
@@ -0,0 +1,23 @@
...
+            [checked]="annotation.display"
+            [title]="annotation.display ? 'Show this annotation on the chart' : 'Hide this annotation on the chart'"
+            color="primary"
+            id="annotationDisplayCheckBox"
+          ></mat-checkbox>
+          {{ annotation.label.content }}
+        </mat-panel-title>
+        <mat-panel-description> Annotation </mat-panel-description>
+      </mat-expansion-panel-header>
+      <annotation-options [annotation]="annotation" (onAnnotationsChange)="this.onAnnotationsChange(annotation, $event)"></annotation-options>
```
semicolin, line 20: forgot to update this when changing event name
```suggestion
      <annotation-options [annotation]="annotation" (annotationChange)="this.onAnnotationsChange(annotation, $event)"></annotation-options>
```

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
semicolin: this comment still needs to be addressed. We're not expecting chooseColorsFromPalette to call itself.

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
semicolin: If you resolve a conversation without applying the suggestion, please provide a reason why you disagree with the suggestion.

semicolin (changes requested):
Nice work. The tests in your most recent commit look good aside from minor issues with variable name and description.

Let me know when you've addressed the other unresolved comments and I'll do another review.

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  it('form value should update from form changes', fakeAsync(() => {
+    let annotationOptions = { color: 'color', label: 'label', yMax: 80, yMin: 120 };
+    updateForm(annotationOptions);
+    fixture.detectChanges();
+    fixture.whenStable().then(() => {
+      expect(component.form.value).toEqual(annotationOptions);
+    });
+  }));
```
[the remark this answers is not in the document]
amolht: ok we have added proper test cases for forms

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-list/annotation-list.component.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+  });
+
+  describe('onAnnotationsChange', () => {
+    it('should update view, control and emit the option if an option is clicked', () => {
+      spyOn(component.annotationsChange, 'emit');
+      const oldAnnotation = { label: { display: true } };
+      const newAnnotation = { label: { display: false } };
+      component.annotations = [oldAnnotation];
+      component.onAnnotationsChange(oldAnnotation, newAnnotation);
+      expect(component.annotationsChange.emit).toHaveBeenCalled();
```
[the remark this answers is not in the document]
amolht: ok I have updated and used real annotation instead of spying

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.ts
```diff
@@ -36,7 +37,7 @@ export class AnnotationOptionsComponent implements OnInit {
 
   private updateModel(formValue: typeof this.form.value): void {
     if (this._annotation) {
-      const props: any = {
+      this.newAnnotation = {
```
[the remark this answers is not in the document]
amolht: yes, previously I thought we need to check these props changes in the test on the event. I updated this

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.spec.ts
semicolin: this test is still checking the FormControl value instead of the annotationChange output

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.spec.ts
semicolin: @amolht looks like code is not updated. do you have another commit that didn't get pushed?

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    const spyColor = colorService.getColor(dataset);
+    spyOn(colorService, 'getColor').and.callFake(function () {
+      if (typeof color === 'string') {
+        return color;
+      } else {
+        return undefined;
+      }
+    });
+    expect(spyColor).toEqual('#e41a1c');
+  });
```
semicolin, line 71: I think this test can be removed. It's doing the exact same thing as the one above. The spy is unused.

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+  describe('setColor', () => {
+    it('should setColor function called', inject([DataLayerColorService], (service: DataLayerColorService) => {
+      const dataset: any = { borderColor: '#e41a1c' };
+      service.setColor(dataset, '#e41a1c');
+      expect('#e41a1c').toEqual(dataset.borderColor);
```
semicolin, line 76:
```suggestion
      expect(dataset.borderColor).toEqual('#e41a1c');
```

projects/showcase/src/app/datasets/blood-pressure-mapper.ts
```diff
@@ -38,12 +38,15 @@ export class BloodPressureMapper implements Mapper<BloodPressureObservation> {
       //   yMin: 80,
       // }),
       merge({}, this.annotationOptions, {
+        display: true,
```
semicolin, line 41: is `display: true` necessary?

semicolin (changes requested):
Almost there. I noticed a few issues from resolved/outdated comments are not fixed. Do you have more changes to push?

projects/showcase/src/app/datasets/blood-pressure-mapper.ts
amolht: yes, colin it is necessary. we are using that display property in the checkbox checked, I have tested by removing this but then it not working properly.

projects/ngx-charts-on-fhir/src/lib/data-layer-list/annotation-options/annotation-options.component.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+    await yMaxInputHarness.setValue('100');
+    expect(yMaxInputControl.value).toBe(100);
+  });
+
+  it('should update annotation yMin when input is changed', async () => {
+    let yMinInputHarness = await loader.getHarness(MatInputHarness.with({ selector: "[id='yMin']" }));
+    let yMinInputControl = fixture.componentInstance.form.controls['yMin'];
+    await yMinInputHarness.setValue('100');
+    expect(yMinInputControl.value).toBe(100);
+  });
```
semicolin, line 80: these two tests still need to be updated to check the emitted event instead of FormControl value

semicolin (changes requested):

semicolin (approved):
Looks good!

yashpmohod (approved):

PR #14, opened 2022-12-09
added category scale for medication mapper

semicolin:
I got this working with a line/scatter chart that uses strings as the y-values. I can't get bar chart to display properly because it seems to choose an arbitrary starting point for the bar based on the initial zoom.

### DIFF
--- a/projects/ngx-charts-on-fhir/src/lib/fhir-mappers/simple-medication-mapper.service.ts
+++ b/projects/ngx-charts-on-fhir/src/lib/fhir-mappers/simple-medication-mapper.service.ts
@@ -4,11 +4,10 @@ import { MedicationOrder } from 'fhir/r2';
 import { merge } from 'lodash-es';
 import { DataLayer } from '../data-layer/data-layer';
 import { Mapper } from '../fhir-converter/multi-mapper.service';
-import { ChartAnnotation } from '../utils';
-import { TIME_SCALE_OPTIONS, LINEAR_SCALE_OPTIONS, ANNOTATION_OPTIONS } from './fhir-mapper-options';
+import { TIME_SCALE_OPTIONS, CATEGORY_SCALE_OPTIONS } from './fhir-mapper-options';
 import { FhirMappersModule } from './fhir-mappers.module';
 
-/** Required properties for mapping an Observation with [SimpleMedicationMapper] */
+/** Required properties for mapping a MedicationOrder with [SimpleMedicationMapper] */
 export type SimpleMedication = {
   medicationCodeableConcept: {
     text: string;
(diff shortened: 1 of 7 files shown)


projects/ngx-charts-on-fhir/src/lib/fhir-mappers/simple-medication-mapper.service.ts
```diff
@@ -4,72 +4,53 @@ import { MedicationOrder } from 'fhir/r2';
...
 } & MedicationOrder;
 export function isMedication(resource: MedicationOrder): resource is SimpleMedication {
-  return !!(
-    resource.resourceType === 'MedicationOrder' &&
-    resource.prescriber?.reference &&
-    resource.dosageInstruction &&
-    resource.medicationCodeableConcept
-  );
+  return !!(resource.resourceType === 'MedicationOrder' && resource.dateWritten && resource.medicationCodeableConcept?.text);
 }
```
semicolin, line 19: This function is a type guard for the parameter type that is passed into the `SimpleMedicationMapper#map` function. It's important that the logic here correctly discriminates the `SimpleMedication` type so that we don't get runtime type errors for resources with missing properties (and also to keep the mapping code simple). For example, if the `SimpleMedication` type asserts that `dateWritten` will always be present, then the `map` function can use it without sprinkling null checks everywhere, but this means the type guard (`canMap` function) needs to return false if `dateWritten` is undefined.

I also removed the extra properties that were not used by the mapper. If properties are not relevant, there's no need to check them in the type guard (unless of course you **want** to filter out those resources).

projects/ngx-charts-on-fhir/src/lib/fhir-mappers/simple-medication-mapper.service.ts
```diff
@@ -4,72 +4,53 @@ import { MedicationOrder } from 'fhir/r2';
...
+    @Inject(CATEGORY_SCALE_OPTIONS) private categoryScaleOptions: ScaleOptions<'category'>
   ) {}
   canMap = isMedication;
-//   created Medication Data Layer 
-  map(resource: MedicationOrder): any {
-    let endDate = new Date(resource.dateWritten ? resource.dateWritten : "" )
-    endDate.setDate(endDate.getDate() + 2) 
-    const output =  {
+  map(resource: SimpleMedication): DataLayer {
+    return {
```
semicolin, line 32: I changed the parameter type to make use of the type and type guard that we've defined above.
I also added the return type to make sure we are returning a valid `DataLayer`.

projects/ngx-charts-on-fhir/src/lib/fhir-mappers/simple-medication-mapper.service.ts
```diff
@@ -4,72 +4,53 @@ import { MedicationOrder } from 'fhir/r2';
...
       name: resource?.medicationCodeableConcept?.text,
       category: 'Medication',
       datasets: [
         {
           label: resource?.medicationCodeableConcept?.text,
-        //   yAxisID: resource.valueQuantity.unit,
+          yAxisID: 'medications',
```
semicolin, line 38: `yAxisId` should match a key in the `scales` object (see below)

projects/ngx-charts-on-fhir/src/lib/fhir-mappers/simple-medication-mapper.service.ts
```diff
@@ -4,72 +4,53 @@ import { MedicationOrder } from 'fhir/r2';
...
+          indexAxis: 'y',
```
semicolin, line 39: `indexAxis: 'y'` makes bar charts horizontal. I couldn't get bar chart working properly, but I left this in here in case you want to give it a try.

projects/ngx-charts-on-fhir/src/lib/fhir-mappers/simple-medication-mapper.service.ts
```diff
@@ -4,72 +4,53 @@ import { MedicationOrder } from 'fhir/r2';
...
           data: [
-            [
-              new Date(resource.dateWritten ? resource.dateWritten : "" ),
-              endDate
-            ],
+            {
+              x: new Date(resource.dateWritten).getTime(),
+              y: resource?.medicationCodeableConcept?.text,
+            },
           ],
```
semicolin, line 45: we can now use a string as y-value because the new `TimelineDataPoint` type is default data type param for `DataLayer`

projects/ngx-charts-on-fhir/src/lib/fhir-mappers/simple-medication-mapper.service.ts
```diff
@@ -4,72 +4,53 @@ import { MedicationOrder } from 'fhir/r2';
...
-    //       yMax: range?.high?.value,
-    //       yMin: range?.low?.value,
-    //     })
-    //   ),
+      scales: {
+        timeline: this.timeScaleOptions,
+        medications: merge({}, this.categoryScaleOptions, {
+          title: { text: 'Medications' },
+        }),
+      },
```
semicolin, line 53: Here is the scale configuration for category axis. it uses a hardcoded key (`medications`) so all medications will share the same scale. Each distinct y-value becomes the label for a tick on the category scale.

All of the examples in chart.js documentation used a `labels` array for category axis, but I found that is not actually necessary if we just strings as y-values. I think this way is much cleaner than using a separate labels array.

yashpmohod (approved):

semicolin:
@yashpmohod feel free to merge these changes into your branch if you want

semicolin:
This has already been merged into another feature branch

PR #17, opened 2022-12-09
Run afterDataLimits callback inside NgZone

semicolin:
## Overview

Change detection was not running for changes made in response to chart.js `afterDataLimits` callback. This bug was not apparent because the Analysis component was manually triggering change detection for its dynamic components. But when I tried using `DataLayerManager#timelineRange$` in another template binding, it was not working.

This PR updates `DataLayerManager` to execute the callback inside NgZone.

## How it was tested

- Created a unit test to verify NgZone.run is getting called.
- Added statistics component directly to app.component, bound its input to timelineRange$, and verified that the component updated when I zoom/pan the chart. (this change to app.component is not included in this PR)

## Checklist

### DIFF
--- a/projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-manager.service.spec.ts
+++ b/projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-manager.service.spec.ts
@@ -3,11 +3,12 @@ import { DataLayerMergeService } from './data-layer-merge.service';
 import { cold, hot, getTestScheduler } from 'jasmine-marbles';
 import { DataLayerColorService } from './data-layer-color.service';
 import { ManagedDataLayer } from './data-layer';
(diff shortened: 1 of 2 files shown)


projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-manager.service.spec.ts
```diff
@@ -100,10 +104,10 @@ describe('DataLayerManagerService', () => {
       );
     });
 
-    it('should emit $timelineRange when timeline scale limits change', waitForAsync(() => {
```
semicolin, line 103: I cleaned up unnecessary waitForAsync, which is not needed with jasmine-marbles test scheduler

amolht (approved):

amolht (approved):

PR #22, opened 2022-12-20
Fix build errors for medication summary

semicolin:
## Overview

- Resolves logical conflicts between medication and chart summary branches.
- `VisibleDataService` now filters out string y-values so `$visible.data` only contains valid `ScatterDataPoint`s
- Statistics panel does not really make sense for medications. It currently displays N/A or NaN for most rows. I will open a new issue for that. This PR only fixes the build errors.

## How it was tested

Added some medications and observations to the chart

## Checklist

### DIFF
--- a/projects/ngx-charts-on-fhir/src/lib/data-layer/visible-data.service.ts
+++ b/projects/ngx-charts-on-fhir/src/lib/data-layer/visible-data.service.ts
@@ -4,7 +4,7 @@ import { OperatorFunction, pipe, map, Observable, combineLatestWith } from 'rxjs
 import { DateRange } from '../analysis/analysis-card-content.component';
 import { ManagedDataLayer, DataLayer, Dataset } from './data-layer';
 import { DataLayerManagerService } from './data-layer-manager.service';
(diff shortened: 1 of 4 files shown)


semicolin:
image

projects/ngx-charts-on-fhir/src/lib/data-layer/visible-data.service.ts
```diff
@@ -45,7 +45,10 @@ function byMostRecentValue(a: Dataset, b: Dataset) {
   if (ay == null || by == null) {
     return 0;
   }
-  return by - ay;
+  if (typeof ay === 'number' && typeof by === 'number') {
+    return by - ay;
+  }
+  throw new TypeError('Unsupported y-value types');
```
semicolin, line 51: I decided not to bother with string comparisons because we do not currently have any DataLayerServices that return multiple category-scaled datasets in the same layer. If we do add such a service in the future, we will probably need to revisit this code anyways. For now, this code never gets called with string values because there is only 1 medication dataset per layer.

yashpmohod (approved):

amolht (approved):

PR #26, opened 2023-01-02
RT-541:Timespan presets

yashpmohod:
## Overview
Added range selector for app
For example: 1 month, 3 months, 6 months, 1 year, and all data.   
And manual date selector for the range 

## How it was tested
Tested by running a app on local 
Added / Run test cases 

## Checklist


yashpmohod:
image_2022_12_29T13_44_49_241Z

amolht (approved):

semicolin (changes requested):

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.ts
```diff
@@ -0,0 +1,81 @@
...
+        let xcordinates = item.map((el: any) => el.x)
+        sortedData = sortedData.concat(xcordinates)
+      }
+      sortedData = sortedData.sort((x: any, y: any) => {
+        return x - y;
+      })
+      this.maxDate = new Date(sortedData[sortedData.length - 1]);
+      this.minDate = new Date(sortedData[0])
+    }
+  }
```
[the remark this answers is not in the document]
yashpmohod: @semicolin, this function is not for resetting zoom,   
Screenshot from 2023-01-04 12-03-31
We have added this for getting max axis form all layer

projects/showcase/src/styles.scss
```diff
@@ -16,6 +18,28 @@ body {
...
+
+  input.mat-input-element {
+    position: relative !important;
+    bottom: 0px !important;
+  }
+
+  .mat-datepicker-toggle {
+    position: relative !important;
+    bottom: 20px !important;
+  }
```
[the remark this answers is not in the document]
yashpmohod: custom styles are needed because if I remove custom styles it's not look in the proper format.
I am attaching a screenshot for you to look over.
Screenshot from 2023-01-04 17-59-52

projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.html
```diff
@@ -1,9 +1,11 @@
 <div class="chart-container" [style.width]="width" [style.height]="height">
   <div class="chart-container-resize-fix">
+    <range-selector></range-selector>
     <canvas
```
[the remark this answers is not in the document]
yashpmohod: Okay, I include that component outside the container div.

projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.spec.ts
```diff
@@ -22,7 +23,7 @@ describe('FhirChartComponent', () => {
 
   beforeEach(async () => {
     await TestBed.configureTestingModule({
-      imports: [FhirChartModule],
+      imports: [FhirChartModule, BrowserAnimationsModule],
```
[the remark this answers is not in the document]
yashpmohod: Yes, I am using NoopAnimationsModule instead of the BrowserAnimationsModule.

projects/showcase/src/styles.scss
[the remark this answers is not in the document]
yashpmohod: If I add custom CSS it looks better and in a proper format.
I am attaching a screenshot for you to take a look at.
image

package.json
```diff
@@ -20,6 +20,8 @@
     "@angular/platform-browser": "^14.2.0",
     "@angular/platform-browser-dynamic": "^14.2.0",
     "@angular/router": "^14.2.0",
+    "@material/button": "^14.0.0",
```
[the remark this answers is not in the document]
yashpmohod: Yes, I have to remove the @material/button.

package.json
```diff
@@ -20,6 +20,8 @@
...
+    "angular-mat-datepicker": "^0.0.2",
```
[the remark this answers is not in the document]
yashpmohod: Yes, I have to remove the angular-mat-date picker.

angular.json
```diff
@@ -124,5 +124,8 @@
         }
       }
     }
+  },
+  "cli": {
+    "analytics": false
```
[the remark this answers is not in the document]
yashpmohod: Okay, I am removing that from angular.json for that moment.

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.ts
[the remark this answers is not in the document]
semicolin: Ok, I see. The maxDate is used in `updateRangeSelector` to calculate the minDate.

You might be able to use [`chart.getInitialScaleBounds()`](https://www.chartjs.org/chartjs-plugin-zoom/latest/guide/developers.html#chart-getinitialscalebounds-record-string-min-number-max-number) to get min and max value of `timeline` scale, which chartjs has already calculated based on all visible data.

projects/showcase/src/styles.scss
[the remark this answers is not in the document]
semicolin: Thanks, I will take a look

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.ts
```diff
@@ -0,0 +1,81 @@
...
+
+@Component({
+  selector: 'range-selector',
+  templateUrl: './range-selector.component.html',
+  styleUrls: ['./range-selector.component.css'],
+})
+export class RangeSelectorComponent {
+  layers?: any[];
+  maxDate: any
+  minDate: any
```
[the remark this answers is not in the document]
yashpmohod: we need to keep the type of these properties as any Because the maxDate should be the date, string, or undefined same for minDate.

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.ts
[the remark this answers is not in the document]
yashpmohod: I am trying to use this chart.getInitialScaleBounds() but sometimes we get the min value of the timeline Scale is 0.
image

And only for one graph does it returns undefined.
image

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.ts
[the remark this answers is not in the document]
semicolin: then use a [Union Type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types)
```ts
  maxDate: Date | string | undefined;
```

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.ts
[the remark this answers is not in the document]
semicolin: hmmm, ok, probably something to do with the timing. We can stick with your solution for now.

projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
```diff
@@ -26,7 +26,9 @@ export class FhirChartComponent implements OnInit {
   @Input() width: string = '600px';
   @Input() height: string = '300px';
 
-  constructor(private configService: FhirChartConfigurationService) {}
+  @ViewChild('baseChart') baseChart?:ElementRef;
```
semicolin, line 29: Is this necessary? I don't see anything in the ts code using this. If it's only used in the template, can't it just use the template reference variable directly?

projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
```diff
@@ -65,4 +67,7 @@ export class FhirChartComponent implements OnInit {
       this.options = config.options;
     });
   }
+  ngAfterContentChecked(): void {
+    this.changeDetector.detectChanges();
+  }
```
semicolin, line 72: why is this needed?

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.spec.ts
```diff
@@ -80,14 +81,10 @@ describe('RangeSelectorComponent', () => {
...
-      name: 'Blood Pressure',
-      category: 'vital-signs',
-      annotations: [],
-      selected: true,
-      enabled: true,
-    }];
+    mockLayerManager.selectedLayers$.subscribe((layers) => {
+      component.getMaxDateFromLayers(layers);
+      component.layers = layers as DataLayer[];
+    })
```
semicolin, line 87: why is this test subscribing to a mock observable?

projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
yashpmohod: @semicolin i have tried this but getting error 
image
Do you have any other solution?

projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
yashpmohod: I was getting this error when I am passing canvas reference to a child, I search for this error and we got a solution we add this because the reference is  undefined, and then it will be defined, I also manually check that but then zoom is not working 
image

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.spec.ts
yashpmohod: We are checking when one or more layer is selected then and only then do we need to display the range selector so in my mockLayerManager I have selected layer I am getting that and storing in  component.layer
Yes we do not need  `component.getMaxDateFromLayers(layers)`  here

projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
semicolin: Running an extra change detection cycle doesn't seem like the best solution to `ExpressionChangedAfterItHasBeenCheckedError`. I'm guessing this has something to do with passing the component reference around. Maybe we could use a service to share the chart object instead?

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.spec.ts
semicolin: Yes, but the real component code is already subscribing to `selectedLayers$` and setting its own `layers` property. I don't see why the test code needs to duplicate this.

Generally, tests do not make direct calls to their own mocks. Mocks are provided to the real code in place of their real dependencies. The real code should call the mock (if it doesn't, then you don't need a mock).

projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
semicolin: Actually, after seeing the change detector error below, let's try using a service instead.

semicolin: Or you could just go back to using a hard-coded id for now. Keep it simple until we actually need something more sophisticated.

projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
```diff
@@ -26,6 +26,8 @@ export class FhirChartComponent implements OnInit {
   @Input() width: string = '600px';
   @Input() height: string = '300px';
 
+  @ViewChild('baseChart') baseChart?:ElementRef;
```
semicolin, line 29: Is the @ViewChild still needed?

semicolin (changes requested):
I see a lot of errors in the console when running unit tests.

Can you add some tests to verify that `getMaxDateFromLayers` is working correctly? That is the most complicated piece of code and should definitely have test coverage.

projects/showcase/src/styles.scss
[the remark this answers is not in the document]
yashpmohod: Okay, I am changing matLconSuffix to matSuffix and Also removing &nbsp.

angular.json
```diff
@@ -141,5 +141,8 @@
...
```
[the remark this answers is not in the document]
yashpmohod: I am not committing this on git, I keep it locally

projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
yashpmohod: No,@ViewChild is not needed Now. I will remove that.

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
+  it('should display the range selector when one or more data layer is selected', () => {
+    const cards = element.query(By.css(".range-selector"));
+    expect(cards).toBeTruthy();
+  });
+
+  it('should not display the range selector when one or more data layer is selected', () => {
```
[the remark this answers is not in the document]
yashpmohod: Okay, I am changing one or more data to no data.

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.spec.ts
```diff
@@ -141,18 +154,28 @@ describe('RangeSelectorComponent', () => {
...
   });
 
+  it('should check maxDate and minDate', async () => {
+    component.getMaxDateFromLayers(component.layers)
+    const componentMindate = new Date(component.minDate).getTime()
+    const componentMaxdate = new Date(component.maxDate).getTime()
+    if (component.layers) {
+      expect(componentMaxdate).toEqual(component.layers[0].datasets[0].data[1].x);
+      expect(componentMindate).toEqual(component.layers[2].datasets[0].data[1].x);
+    }
```
semicolin, line 178: This test will pass if component.layers is undefined. I don't think that's what we want. 
Try this instead:

```suggestion
    expect(componentMaxdate).toEqual(component.layers?.[0].datasets[0].data[1].x);
    expect(componentMindate).toEqual(component.layers?.[2].datasets[0].data[1].x);
```

projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
```diff
@@ -1,4 +1,4 @@
-import { Component, Input, OnInit } from '@angular/core';
+import { Component, ElementRef, Input, OnInit, ViewChild } from '@angular/core';
```
semicolin, line 1: unused imports?

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.ts
```diff
@@ -0,0 +1,79 @@
...
+  templateUrl: './range-selector.component.html',
+  styleUrls: ['./range-selector.component.css'],
+})
+export class RangeSelectorComponent {
+  layers?: DataLayer[];
+  maxDate: Date | string;
+  minDate: Date | string;
+  isMatGroupFocus : boolean = true;
+
+ constructor(private layerManager: DataLayerManagerService, private el: ElementRef) {
```
semicolin, line 19: is the ElementRef still needed? I don't see it used anywhere

semicolin (changes requested):
A few minor changes...

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.spec.ts
semicolin: ...although that will pass if both the layers and maxDate are undefined. We should probably use a literal date for expected value instead of pulling it from the component. Something like this:
```
expect(componentMaxdate).toEqual(new Date(1573505702))
```

semicolin:
> The start date and end date do not update when zooming/panning the chart

We can create a separate PR for this if you prefer.

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.spec.ts
yashpmohod: @semicolin here we need to typecast that number for compile 
image

projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
yashpmohod: Okay, I will remove this.

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.ts
yashpmohod: No, ElementRef is not needed Now.

PR #27, opened 2023-01-03
RT-552:Annotation options displays the wrong color

yashpmohod:
## Overview
 Fixed the issue 

## How it was tested
Tested on local by running app on local.
 

## Checklist

### DIFF
--- a/projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
+++ b/projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
@@ -29,11 +29,8 @@ describe('DataLayerColorService', () => {
   });
   describe('getAnnotationColor', () => {
     it('should return the annotation background color', inject([DataLayerColorService], (service: DataLayerColorService) => {
-      const annotation: any = {
-        label: {},
-        backgroundColor: '#ECF0F933',
-      };
-      expect(service.getAnnotationColor(annotation)).toEqual(annotation.backgroundColor);
+      const color = '#ECF0F9';
+      expect(service.getAnnotationColor(color)).toEqual('#ECF0F933');
     }));
   });
 
--- a/projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
+++ b/projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
@@ -27,10 +27,10 @@ describe('DataLayerColorService', () => {
       expect(setColorSpy).toHaveBeenCalled();
     }));
   });
-  describe('getAnnotationColor', () => {
+  describe('addTransparencyAbritraryColor', () => {
     it('should return the annotation background color', inject([DataLayerColorService], (service: DataLayerColorService) => {
       const color = '#ECF0F9';
-      expect(service.getAnnotationColor(color)).toEqual('#ECF0F933');
+      expect(service.addTransparencyAbritraryColor(color)).toEqual('#ECF0F933');
     }));
   });
 
--- a/projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
+++ b/projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
@@ -27,10 +27,10 @@ describe('DataLayerColorService', () => {
       expect(setColorSpy).toHaveBeenCalled();
     }));
   });
-  describe('addTransparencyAbritraryColor', () => {
-    it('should return the annotation background color', inject([DataLayerColorService], (service: DataLayerColorService) => {
+  describe('addTransparency', () => {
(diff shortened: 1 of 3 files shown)


projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.ts
```diff
@@ -27,8 +27,7 @@ export class DataLayerColorService {
     line.backgroundColor = color + '33'; // temporary dirty hack to set opacity. assumes color is in 6-character hex format.
   }
 
-  getAnnotationColor(annotation: any): string | undefined {
-    const color = annotation.backgroundColor;
+  getAnnotationColor(color: any): string | undefined {
     if (typeof color === 'string' && !color.includes('33')) {
       return color + '33';
     }
```
semicolin, line 33: I find this confusing that the `getAnnotationColor` method no longer gets the value from the annotation. Based on the name of the method, I would expect it to do something similar to `getColor`. It looks like this is now a utility method for adding transparency to an abritrary color code, so let's change its name to describe what it is doing.

There's also a bug in this method because `includes` will match a '33' anywhere in the string, not just at the end.

semicolin (changes requested):

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.ts
yashpmohod: right, @semicolin i will change it in the component, we don't need a service method for that add a '33' in color

yashpmohod: instead of includes we use endsWith() for checking whether '33' is present in end or not.
And function name changed from getAnnotationColor to addTransparencyAbritraryColor.

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.ts
```diff
@@ -27,9 +27,8 @@ export class DataLayerColorService {
     line.backgroundColor = color + '33'; // temporary dirty hack to set opacity. assumes color is in 6-character hex format.
   }
 
-  getAnnotationColor(annotation: any): string | undefined {
-    const color = annotation.backgroundColor;
-    if (typeof color === 'string' && !color.includes('33')) {
+  addTransparencyAbritraryColor(color: any): string | undefined {
```
semicolin, line 30: Can we shorten this method name to `addTransparency`?

Also, please specify a type for the parameter.

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
```diff
@@ -27,13 +27,10 @@ describe('DataLayerColorService', () => {
       expect(setColorSpy).toHaveBeenCalled();
     }));
   });
-  describe('getAnnotationColor', () => {
+  describe('addTransparencyAbritraryColor', () => {
     it('should return the annotation background color', inject([DataLayerColorService], (service: DataLayerColorService) => {
```
semicolin, line 31: please update this test description to match the new behavior

semicolin (changes requested):
Just a couple minor issues...

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.ts
yashpmohod: Okay, I am changing this method name from addTransparencyAbritraryColor to addTransparency.

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
yashpmohod: Yes, I am updating this test description to match the new behavior.

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.ts
```diff
@@ -27,9 +27,8 @@ export class DataLayerColorService {
     line.backgroundColor = color + '33'; // temporary dirty hack to set opacity. assumes color is in 6-character hex format.
   }
 
-  getAnnotationColor(annotation: any): string | undefined {
-    const color = annotation.backgroundColor;
-    if (typeof color === 'string' && !color.includes('33')) {
+  addTransparency(color: string | undefined): string | undefined {
+    if (typeof color === 'string' && !color.endsWith('33')) {
```
semicolin, line 31: This will not work correctly for colors that end with `33` but have no alpha channel (like `#FFFF33`) or colors that have an alpha channel other than `33` (like `#FFFFFF66`) or colors that use 3-digit hex or rgb or any other type of color string. We really should be using a library for color manipulation, so I'll just approve this PR and create a new issue.

semicolin (approved):

amolht (approved):

PR #30, opened 2023-01-09
update to angular 15

yashpmohod:
## Overview
  Updated a angular version 14 to 15 

## How it was tested
  Tested on local and it working

## Checklist


semicolin:
I see a few visual gitches caused by the [MDC migration](https://rc.material.angular.io/guide/mdc-migration).

1. Toolbar buttons are off-center and too close to the edge because the icon button width was changed from 40px to 48px. Increase the width of the toolbar or use the new density setting (woohoo!) to reduce the button size.
2. The selected color in the picker component is mis-aligned. Not sure why.
3. The chart has no padding. Put a `<mat-card-content>` around the `<fhir-chart>` in app component.
4. The summary card title has no padding. Put a `<mat-card-header>` around the `<mat-card-title-group>` in analysis-card component.

image

semicolin:
I think we can also reduce or remove the margin-right on checkboxes now that they have built-in padding. (data-layer-list, annotation-list, and dataset-list)

semicolin:
The bottom border needs to be removed from the last row for custom `.row-header` style in statistics component:
image

This can be done with a `:last-child` selector:
```css
tr:last-child .row-header {
  border-bottom: none;
}
```

The padding-left on `.mat-column-value` can also be removed because all table cells now have 16px padding by default.

semicolin (changes requested):
See comments above

semicolin:
> Using providedIn: ngModule for an @Injectable and InjectionToken is deprecated in v15

Some of the services in ngx-charts-on-fhir are using this. I will look into changing those in a separate PR.

semicolin:
I think something went wrong with that merge commit. It does not include the changes to tsconfig. You may want to reset and try the merge again.

semicolin:
@yashpmohod most of the glitches that I pointed out above are still present

projects/ngx-charts-on-fhir/src/lib/color-picker/color-picker.component.css
```diff
@@ -6,3 +6,8 @@ button {
   border-radius: 4px;
   max-width: 300px;
 }
+
+.mdc-fab::before {
+  position: relative;
+  display: contents;
+}
```
semicolin, line 13: can you explain why this fixes the issue?

projects/ngx-charts-on-fhir/src/lib/data-layer-toolbar/data-layer-toolbar/data-layer-toolbar.component.css
```diff
@@ -6,3 +6,8 @@
   width: 48px;
   padding-top: 16px;
 }
+
+#badge-movement ::ng-deep .mat-mdc-tooltip-trigger {
+  padding: 12px 15px 12px 4px;
+}
```
semicolin, line 12: I don't think this is the best solution. The buttons are still bigger than the toolbar so the ripple animation overlaps the edge of the toolbar and is not centered on the icon. Can't we just increase the toolbar width so we don't need to modify the material component internals?

semicolin (changes requested):
Still a few issues. And please double-check every merge conflict resolution.

PR #37, opened 2023-01-13
RT-554:Summary card title is too long

yashpmohod:
## Overview
Add an ellipse and max-width for the summary card title, dataset label, and for layer name.
also, add a tooltip on hover to show the layer's title.

## How it was tested
Tested it by the running showcase app locally.
Run unit test cases on local

## Checklist

- [ ] I have included screenshots for changes that affect the user interface


semicolin:
why does this branch have "RT-539 Updated angular version to 15" in the commit history? I don't see any of the angular version changes here. Was that just a mis-labelled commit?

I'm really confused by all those merge commits. What happened with this branch? Why was it merged into itself?

projects/ngx-charts-on-fhir/src/lib/analysis-cards/analysis-card/analysis-card.component.css
```diff
@@ -7,4 +7,9 @@
 }
 .mat-card-title {
   font-size: 20px;
+  max-width: 250px;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  cursor: pointer;
```
semicolin, line 14: `cursor: pointer` should only be used for clickable elements. The card title does nothing when you click on it. This will confuse users.

semicolin (changes requested):
The tooltips are redundant when the full text is already displayed. Is there a simple way to show tooltips only when the text is overflowing? If not, maybe we should only put tooltips on the summary card titles and remove them from the data-layer-list and dataset-list?

yashpmohod:
@semicolin sorry, I will clean this PR.

PR #42, opened 2023-01-16
RT-575:End Date does not work for medications

yashpmohod:
## Overview
Check whether the value of the X axis is a number or not and then filter the layer data.

## How it was tested

tested using the run showcase app locally and add observation and then add medication.

## Checklist

### DIFF
--- a/projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.ts
+++ b/projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.ts
@@ -69,13 +69,9 @@ export class RangeSelectorComponent {
       })
       let sortedData: any[] = [];
       for (let item of data) {
-        let xcordinates = item.map(function (el: any) {
-          if (!isNaN(el.x)) {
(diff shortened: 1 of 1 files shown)


projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.ts
```diff
@@ -69,7 +69,14 @@ export class RangeSelectorComponent {
...
       for (let item of data) {
-        let xcordinates = item.map((el: any) => el.x)
+        let xcordinates = item.map(function (el: any) {
+          if (!isNaN(el.x)) {
+            return el.x;
+          }
+        })
+        xcordinates = xcordinates.filter(function( element:any ) {
+          return element !== undefined;
+       });
```
semicolin, line 79: this could be simplified by just filtering out NaN instead of converting it to undefined first

semicolin (approved):
Just one minor suggestion, but if you want to leave it as-is that's fine too.

semicolin (approved):

PR #43, opened 2023-01-16
RT-577:6mo button stays selected after zooming

yashpmohod:
## Overview
We use the onZoomStart on that, we ate removing the focus of the range selector   

## How it was tested
Tested using the run showcase app locally.

## Checklist

### DIFF
--- a/projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
+++ b/projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
@@ -50,7 +50,7 @@ export class FhirChartComponent implements OnInit {
         wheel: { enabled: true },
         drag: { enabled: false },
         pinch: { enabled: true },
-        onZoomComplete: function (myChart: any) {
+        onZoomComplete: function () {
           that.isMatGroupFocus = !that.isMatGroupFocus ;
         }
       },
--- a/projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
+++ b/projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
@@ -28,7 +28,6 @@ export class FhirChartComponent implements OnInit {
   @Input() width: string = '600px';
   @Input() height: string = '300px';
 
-  isMatGroupFocus: boolean = true;
   
   constructor(private configService: FhirChartConfigurationService) {
   }
@@ -41,7 +40,6 @@ export class FhirChartComponent implements OnInit {
   ngOnInit(): void {
     Chart.register(annotationPlugin, zoomPlugin, scaleStackDividerPlugin);
     Chart.register(MedicationScale);
-    const that=this;
     // To responsively resize the chart based on its container size, we must set maintainAspectRatio = false
     Chart.defaults.maintainAspectRatio = false;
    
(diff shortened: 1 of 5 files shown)


projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
```diff
@@ -28,15 +28,18 @@ export class FhirChartComponent implements OnInit {
...
+  isMatGroupFocus: boolean = true;
+  
+  constructor(private configService: FhirChartConfigurationService) {
+  }
 
   ngOnInit(): void {
     Chart.register(annotationPlugin, zoomPlugin, scaleStackDividerPlugin);
     Chart.register(MedicationScale);
-
+    const that=this;
```
semicolin: this is an anti-pattern. use an arrow function instead.

projects/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.ts
```diff
@@ -47,6 +50,9 @@ export class FhirChartComponent implements OnInit {
         wheel: { enabled: true },
         drag: { enabled: false },
         pinch: { enabled: true },
+        onZoomComplete: function () {
+          that.isMatGroupFocus = !that.isMatGroupFocus ;
+        }
```
semicolin: why do we need to listen to this event? range-selector is already getting the min/max from `layerManager.timelineRange$` every time the scale bounds change.

semicolin: Also, why is it toggling the state on each zoom? is that really the right behavior?

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.ts
```diff
@@ -14,11 +14,16 @@ export class RangeSelectorComponent {
   layers?: DataLayer[];
   maxDate: Date | string;
   minDate: Date | string;
-  isMatGroupFocus: boolean = true;
-
+  _isMatGroupFocus: boolean = true;
+  @Input() 
+  set isMatGroupFocus(isMatGroupFocus: boolean){
+    this._isMatGroupFocus = isMatGroupFocus;
+  }
```
semicolin: why do we need a setter for this instead of a simple property? the setter is not doing anything except setting the value.

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.ts
```diff
@@ -32,6 +37,11 @@ export class RangeSelectorComponent {
     })
 
   }
+  ngOnChanges(changes: SimpleChanges){
+    const changedProp = changes["isMatGroupFocus"];
+    this._isMatGroupFocus = changedProp.currentValue;
+  }
```
semicolin: it looks like this is doing the exact same thing as the setter? why do we need so much complexity and duplication just to set a property value?

semicolin (changes requested):
I recommend deleting this branch and starting over. It shouldn't be this complicated.
The range-selector component already knows the minDate and maxDate. All you need is a function that converts minDate/maxDate into a "onemonth" | "threemonth" | etc. so you can bind it to the mat-button-toggle-group's `[value]`

semicolin (changes requested):

yashpmohod:
we have created another PR for the RT-577 branch.

PR #48, opened 2023-01-18
Use fixed scale bounds

semicolin:
## Overview

- Set the min/max for each scale based on the min/max values of the datasets and annotations.
- This prevents the scale ticks from changing when zooming/panning the chart.

## How it was tested

- Ran showcase app with mock FHIR server
- Added many different layers (observations and medications)
- Verified that all datapoints and annotations are visible
- Zoomed and panned to check that scale bounds do not change

## Checklist


dimon37 (approved):
Looks reasonable

PR #65, opened 2023-01-23
Create GitHub workflow for deployment

semicolin:
## Overview

- Use webhook URL to trigger deployment on push to main

## How it was tested

- Cannot test until it is merged

## Checklist


semicolin:
I think this would run in parallel with build-test workflow, but we probably want to require successful build & test before deploying. I will create another PR that adds a job to the build-test workflow instead.

PR #67, opened 2023-01-24
Get clientId from env

semicolin:
## Overview

- Moved clientId from launch.html to assets/env.js file
- Added env.template.js file with env var for clientId
- Dockerfile runs `envsubst` to generate env.js by replacing env vars in env.template.js when the container is started
- Copy package-lock.json and use `npm ci`

## How it was tested

- Build & run docker container
- Inspected the served env.js to make sure it had the clientId from env var
- Tested SMART launch from Logica sandbox
- Regression tested `ng serve` without docker

## Checklist


Dockerfile
```diff
@@ -15,4 +15,5 @@ RUN rm -rf /usr/share/nginx/html/*
 COPY nginx.conf /etc/nginx/
 ARG project
 COPY --from=build /home/node/dist/${project} /usr/share/nginx/html
-CMD ["nginx"]
+# envsubst is used to replace environment variables in env.js when the container is started
+CMD ["/bin/sh", "-c", "envsubst < /usr/share/nginx/html/assets/env.template.js > /usr/share/nginx/html/assets/env.js && exec nginx"]
```
hankwallace, line 19: Consider creating a `docker-entrypoint.sh` script that is executed here instead. See one of our internal repos like the sidecars for examples.

It should be something like:
```
ENTRYPOINT ["/etc/nginx/docker-entrypoint.sh"]
CMD ["nginx"]
```

projects/showcase/src/assets/env.js
```diff
@@ -0,0 +1,5 @@
+// This client ID is registered in Logica Sandbox for local development (http://localhost:4200).
+// In production builds, this file should be generated by passing env.template.js into `envsusbt` during container startup
+window.env = {
+  clientId: "2dd5083f-85ba-4966-8b04-3c99dcd26601",
```
hankwallace, line 4: Is this a secret or just a sample UUID?

Dockerfile
hankwallace: What happens if this is started without a `CLIENT_ID`? Should there be a `ENV` statement with a default here?

hankwallace (changes requested):

Dockerfile
semicolin: I don't think a default CLIENT_ID would make sense. The app will not work unless the client ID is registered with the EHR using the correct launch URL. The localhost client ID will not work when the app is deployed at a different URL. I will add a check in the docker-entrypoint.sh and fail if the env is not configured correctly.

semicolin: The app could be configured to connect to an open FHIR server without SMART launch. In that case, we would want to allow starting the container without a CLIENT_ID defined. However, we do not currently have any way to select a patient when using an open FHIR server. So until/unless we add that feature, I think it still makes sense to fail when the env var is not set. At this stage it is more likely to catch a configuration mistake than to prevent something intentional.

hankwallace: Ok, that makes sense.

projects/showcase/src/assets/env.js
hankwallace: Ok, understood.

Dockerfile
hankwallace: Please change the `CMD` and `ENTRYPOINT` though.

hankwallace (approved):

PR #75, opened 2023-01-25
Fix timeline resetting when options are changed

semicolin:
## Overview

- Timeline range is now locked when the user zooms, pans, or clicks the range-selector buttons.
- The "All" button unlocks the zoom and restores the auto-zooming behavior.
- Fixed various issues related to zooming and calculating data bounds.
- Refactored range-selector, moving most of the zoom/bounds code into `FhirChartConfigurationService`
- Code cleanup

## How it was tested

- Ran Showcase app with mock FHIR server
- Added observation and medication layers in various combinations
- Tested hiding/showing each layer with each preset button selected
- Tested changing color and interpolation settings with each preset button selected
- Tested zoom/pan by clicking in the chart to make sure it locks zoom when layer is added/removed
- Tested that preset buttons select/deselect appropriately when zooming/panning the chart. The "All" button is a bit unintuitive because if you zoom out it deselects even though all of the data is still visible. We might want to rename it to "Auto"?

## Checklist


semicolin:
image

dimon37 (approved):
Looks reasonable.

PR #85, opened 2023-02-01
RT-633-Summary-Card:Summary Card is not removed for disabled layers

yashpmohod:
## Overview
Add the ngIf condition to check layer is enabled or not.

image

## How it was tested
Tested using run showcase app locally.

## Checklist

### DIFF
--- a/projects/ngx-charts-on-fhir/src/lib/fhir-chart-summary/fhir-chart-summary/fhir-chart-summary.component.html
+++ b/projects/ngx-charts-on-fhir/src/lib/fhir-chart-summary/fhir-chart-summary/fhir-chart-summary.component.html
@@ -5,7 +5,8 @@
         [title]="layer.name" 
         icon="info"
         class="dataset-summary mat-elevation-z2">
-        <dynamic-table [data]="summarize(layer, range)"></dynamic-table>
+        <dynamic-table [data]="summarize(layer, range)">
+        </dynamic-table>
       </fhir-chart-summary-card>
   </ng-container>
 </div>
--- a/projects/ngx-charts-on-fhir/src/lib/fhir-chart-summary/fhir-chart-summary/fhir-chart-summary.component.html
+++ b/projects/ngx-charts-on-fhir/src/lib/fhir-chart-summary/fhir-chart-summary/fhir-chart-summary.component.html
@@ -1,13 +1,13 @@
 <div class="chart-summary">
   <ng-container *ngIf="configService.timelineRange$ | async as range">
-    <fhir-chart-summary-card
-      *ngFor="let layer of layerManager.selectedLayers$ | async"
-      [color]="colorService.getColorGradient(layer)"
-      [title]="layer.name"
-      icon="info"
-      class="dataset-summary mat-elevation-z2"
-    >
-      <dynamic-table [data]="summarize(layer, range)"></dynamic-table>
-    </fhir-chart-summary-card>
+    <div class="summary-card" *ngFor="let layer of layerManager.selectedLayers$ | async">
+      <fhir-chart-summary-card  *ngIf="layer.enabled"
+        [color]="colorService.getColorGradient(layer)" 
+        [title]="layer.name" 
+        icon="info"
+        class="dataset-summary mat-elevation-z2">
+        <dynamic-table [data]="summarize(layer, range)"></dynamic-table>
+      </fhir-chart-summary-card>
+    </div>
   </ng-container>
 </div>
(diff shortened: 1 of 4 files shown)


projects/ngx-charts-on-fhir/src/lib/fhir-chart-summary/fhir-chart-summary/fhir-chart-summary.component.html
```diff
@@ -1,13 +1,13 @@
...
-      *ngFor="let layer of layerManager.selectedLayers$ | async"
-      [color]="colorService.getColorGradient(layer)"
-      [title]="layer.name"
-      icon="info"
-      class="dataset-summary mat-elevation-z2"
-    >
-      <dynamic-table [data]="summarize(layer, range)"></dynamic-table>
-    </fhir-chart-summary-card>
+    <div class="summary-card" *ngFor="let layer of layerManager.selectedLayers$ | async">
+      <fhir-chart-summary-card  *ngIf="layer.enabled"
```
semicolin, line 4: This leaves an empty div for each hidden layer. It's barely noticeable with one hidden layer, but with more layers it can become a large gap.
image

projects/ngx-charts-on-fhir/src/lib/fhir-chart-summary/fhir-chart-summary/fhir-chart-summary.component.spec.ts
```diff
@@ -67,20 +67,20 @@ describe('FhirChartSummaryComponent', () => {
     const layer: ManagedDataLayer = { id: '1', name: 'layer', datasets: [], scale: { id: 'test' } };
     layerManager.selectedLayers$.next([layer, layer, layer]);
     fixture.detectChanges();
-    const cards = fixture.debugElement.queryAll(By.directive(MockFhirChartSummaryCardComponent));
+    const cards = fixture.debugElement.queryAll(By.css('.summary-card'));
```
semicolin, line 70: Why did you change this? we still want to make the sure the cards are rendered and not just empty divs, right?
Also, please add a test for `it('should not render a card for disabled layers')`

projects/ngx-charts-on-fhir/src/lib/fhir-chart-summary/fhir-chart-summary/fhir-chart-summary.component.html
semicolin: You can fix this either by using `<ng-container>` instead of `<div>`
or by creating an observable like:
```
  enabledLayers$ = this.selectedLayers$.pipe(map((layers) => layers.filter((layer) => layer.enabled)));
```
which might be useful in other places too.

semicolin (changes requested):

amolht (dismissed):

projects/ngx-charts-on-fhir/src/lib/fhir-chart-summary/fhir-chart-summary/fhir-chart-summary.component.spec.ts
```diff
@@ -65,25 +66,36 @@ describe('FhirChartSummaryComponent', () => {
...
     const statistics: DebugElement = fixture.debugElement.query(By.directive(MockDynamicTableComponent));
     expect(statistics.componentInstance.data).toEqual([{ name: 'summary' }]);
   });
+  it('should not render a card for disabled layers', () => {
+    const enabledLayer: ManagedDataLayer = { id: '1', name: 'layer', datasets: [], scale: { id: 'test' },enabled:true };
+    const disabledLayer: ManagedDataLayer = { id: '1', name: 'layer', datasets: [], scale: { id: 'test' },enabled:false };
+    layerManager.selectedLayers$.next([enabledLayer, disabledLayer]);
+    layerManager.selectedLayers$.pipe(map((layers) => layers.filter((layer) => layer.enabled))).forEach(layer => {
+      layerManager.enabledLayers$.next(layer)
+    });
```
semicolin, line 96: This is really confusing. The test should not be subscribing to a mock observable. This component no longer depends on selectedLayers$ so I don't understand why the test is mocking that.

It also should not be duplicating the enabledLayers$ logic from the real code. It seems like this test is really only testing itself.

semicolin: Since you moved the logic for checking enabled layers into LayerManagerService, you should probably add an enabledLayers$ test in that class and delete this one.

projects/ngx-charts-on-fhir/src/lib/fhir-chart-summary/fhir-chart-summary/fhir-chart-summary.component.html
```diff
@@ -1,13 +1,11 @@
...
-      <dynamic-table [data]="summarize(layer, range)"></dynamic-table>
-    </fhir-chart-summary-card>
+      <fhir-chart-summary-card *ngFor="let layer of layerManager.enabledLayers$ | async"
+        [color]="colorService.getColorGradient(layer)" 
+        [title]="layer.name" 
+        icon="info"
+        class="dataset-summary mat-elevation-z2">
+        <dynamic-table [data]="summarize(layer, range)"></dynamic-table>
+      </fhir-chart-summary-card>
   </ng-container>
```
semicolin, line 10: please format

semicolin (changes requested):

semicolin (approved):

PR #90, opened 2023-02-03
Rt 631 use fixed layer order

amolht:
## Overview
Added fixed layer order
Order support for both medication and conservation layers  

## How it was tested
Tested on local by running cardiovascular health app on local. 

## Checklist


amolht:
Screenshot from 2023-02-03 18-41-46

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-manager.service.ts
```diff
@@ -62,19 +62,36 @@ export class DataLayerManagerService {
...
+    selected?.map(id => this.stateSubject.next(
+      produce(this.state, (draft) => {
+        const layer = draft.layers[id];
+        console.log(draft.layers)
+        draft.selected.push(layer.id);
+        layer.selected = true;
+        layer.enabled = true;
+        this.colorService.chooseColorsFromPalette(layer);
+      })
+    ))
```
semicolin, line 93: 1. Why does this need to duplicate the code from the select method? can it just call the method instead?
2. If you are not using the return value of `map`, use `forEach` or a `for...of` loop instead.

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-manager.service.ts
```diff
@@ -62,19 +62,36 @@ export class DataLayerManagerService {
...
         this.stateSubject.next({
           ...this.stateSubject.value,
           layers: this.mergeService.merge(this.state.layers, layer),
         }),
+
       error: (err) => console.error(err),
       complete: () => {
         this.loading$.next(false);
+        this.selectAll(selected)
       },
```
semicolin, line 78: So the layers will not show up until all data has finished loading?

semicolin (changes requested):
I don't think we need to support pre-selecting a specific list of layers. We can just auto-select all or none.

semicolin:
> I don't think we need to support pre-selecting a specific list of layers. We can just auto-select all or none.

Sorry, I forgot that the list is also defining the layer order. Can we try using `layer.scale.weight` to define the order?

amolht:
> > I don't think we need to support pre-selecting a specific list of layers. We can just auto-select all or none.
> 
> Sorry, I forgot that the list is also defining the layer order. Can we try using `layer.scale.weight` to define the order?

amolht:
@semicolin, I have tried with `layer.scale.weight` but it only orders chat layers not selected layers and statistic cards.

projects/cardiovascular-health/src/app/app.component.ts
```diff
@@ -10,16 +11,21 @@ import { DataLayerManagerService } from 'ngx-charts-on-fhir';
...
+    this.layerManager.retrieveAll(this.orderLayers, this.isAllLayerSelected);
+  }
 
-    this.layerManager.availableLayers$.subscribe((layers) => {
-      layers.forEach((layer) => this.layerManager.select(layer.id));
-    });
+  orderLayers(layer: any) {
+    const layerOrder: string[] = ['Heart rate', 'Blood Pressure', 'O2 Sat', 'Glucose', 'Step Count', 'Body Weight', 'Medications']
+    return layer.pipe(
+      toArray(),
```
semicolin, line 25: `toArray` waits until the observable competes before emitting anything. I don't think that's what we want here

projects/cardiovascular-health/src/app/app.component.ts
```diff
@@ -10,16 +11,21 @@ import { DataLayerManagerService } from 'ngx-charts-on-fhir';
...
+      map((things: any) => things.sort((a: any, b: any) => layerOrder.indexOf(a.name) - layerOrder.indexOf(b.name))),
+      mergeAll()
```
semicolin, line 27: Why are you using `mergeAll` here? Is this really a higher-order observable? It doesn't look like it based on the declared types in DataLayerManager.

projects/cardiovascular-health/src/app/app.component.ts
```diff
@@ -10,16 +11,21 @@ import { DataLayerManagerService } from 'ngx-charts-on-fhir';
...
 
   ngOnInit(): void {
-    this.layerManager.retrieveAll(this.layerOrder);
+    this.layerManager.retrieveAll(this.orderLayers, this.isAllLayerSelected);
+  }
 
-    this.layerManager.availableLayers$.subscribe((layers) => {
-      layers.forEach((layer) => this.layerManager.select(layer.id));
-    });
+  orderLayers(layer: any) {
```
semicolin, line 22: Please declare the parameter type. Type-checking is very helpful for code like this.

semicolin (changes requested):
Please simplify the API. The type for the `layerOrder` parameter is confusing and I don't understand why the client code needs all those rxjs operators just to implement a sort callback. Take a look at other APIs that use a sort callback to get ideas (e.g. Array.sort or _.sortBy).

semicolin (changes requested):

projects/ngx-charts-on-fhir/src/lib/data-layer/data-layer-manager.service.ts
```diff
@@ -64,23 +65,48 @@ export class DataLayerManagerService {
...
   }
+  
+  /** Reducer that returns a new state with selected layers sorted by `sortCompareFn` */
+  private sortLayers = produce<DataLayerManagerState, [LayerCompareFn]>((draft, sortCompareFn) => {
+    draft.selected.sort((idA, idB) => sortCompareFn(draft.layers[idA], draft.layers[idB]));
+  });
+
+  /** Reducer that returns a new state with all layers selected */
+  private selectAllLayers = (state: DataLayerManagerState) => {
+    return Object.keys(this.state.layers).reduce((nextState, id) => this.selectLayer(nextState, id), state);
```
semicolin, line 97: this should be using layers from the state parameter, not the instance property
```suggestion
    return Object.keys(state.layers).reduce((nextState, id) => this.selectLayer(nextState, id), state);
```

semicolin (changes requested):

semicolin:
We should add unit tests for sort and selectAll

semicolin (approved):

yashpmohod (approved):

PR #91, opened 2023-02-06
RT-629:BP entry form, post data on logica for BP

yashpmohod:
## Overview
Add function in **fhir-data.service.ts**  for post data on logical for BP. 
image
image

image

## How it was tested

Tested using Run showcase and cardiovascular-patient app 

## Checklist


projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.ts
```diff
@@ -71,6 +71,83 @@ export class FhirDataService {
...
+          }
+        ],
+        "text": "Blood Pressure"
+      },
+      "subject": {
+        "reference": `Patient/${this.client?.patient.id}`
+      },
+      "encounter": {
+        "reference": "Encounter/23249"
+      },
```
semicolin, line 104: We should not use a hard-coded encounter ID

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.ts
```diff
@@ -71,6 +71,83 @@ export class FhirDataService {
       return teardownLogic;
     });
   }
+  addPatientData(reportBPValue: any) {
+    const resource = {
```
semicolin, line 75: Can we separate the concerns of  1) creating the Resource object, and 2) making the post request, so that we can reuse the `addPatientData()` function for any kind of data, not just BP?

semicolin (changes requested):
Please add unit tests for new FhirDataService functionality. Also see comments below...

semicolin:
If the security hotspots are all HTTP protocol used in FHIR code system identifiers, you can mark those as safe.

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.ts
yashpmohod: Okay, I will get the encounter ID from client.encounter.

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.ts
yashpmohod: Yes, I will add one more separate function(createResourceData) for creating the Resource Object.

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.ts
```diff
@@ -71,8 +76,15 @@ export class FhirDataService {
       return teardownLogic;
     });
   }
-  addPatientData(reportBPValue: any) {
-    const resource = {
+
+  addPatientData(resource: any, requestOptions: {}) {
```
semicolin, line 80: A parameter of type `{}` (empty object) seems useless. Did you mean to use `{}` as the default value instead?

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.ts
```diff
@@ -100,10 +112,10 @@ export class FhirDataService {
         "reference": `Patient/${this.client?.patient.id}`
       },
       "encounter": {
-        "reference": "Encounter/23249"
+        "reference": `Encounter/${this.client?.encounter.id}`
```
semicolin, line 115: Are you sure we will have an encounter in the context when launching a patient app? We should check if it's there first so we don't add a reference for `Encounter/undefined`

semicolin (changes requested):

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.ts
yashpmohod: okay, I will pass only resource in addPatientData.
image

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.ts
yashpmohod: Yes, We get encounter because we add encounter in environment.
image

semicolin: But that is a fake context that we use for testing. We can put whatever we want in there. The question is whether a real SMART launch for a patient app will include an encounter in the context. We don't need to answer that question here, but we do need to do some defensive coding to ensure that we are not saving FHIR resources that look like this:
```json
"encounter": {
  "reference": "Encounter/undefined"
}
```
Please test your code without an encounter in the context to make sure it still behaves sensibly.

yashpmohod: @semicolin, when I have tried to create a resource without an encounter then it gives 400, and in client, we get null as an encounter

semicolin (changes requested):

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.ts
yashpmohod: I remove the encounter from the resource and now it works without an encounter.

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.ts
```diff
@@ -71,6 +76,91 @@ export class FhirDataService {
...
+
+  addPatientData(resource: any) {
+    if (!this.client) {
+      throw new Error('FhirClientService has not been initialized.');
+    } else {
+      this.client?.create(resource)
+        .then(() => alert('data inserted successfully'))
+        .catch((error) => { return error });
+    }
+  }
```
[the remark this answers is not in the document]
yashpmohod: I am already remove alert() in **RT-624 Configurable success message**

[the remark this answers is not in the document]
semicolin: Why do you have 2 PRs that create different definitions for the same function? This will cause a merge conflict.

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.ts
```diff
@@ -94,6 +99,86 @@ export class FhirDataService {
       return teardownLogic;
     });
   }
+
+  addPatientData(resource: Meta) {
```
semicolin, line 103: I doubt that [Meta](https://hl7.org/fhir/resource.html#Meta) is the correct type to use here. Please take another look at this.

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.spec.ts
```diff
@@ -113,6 +118,16 @@ describe('FhirDataService', () => {
...
 
+  it('shold create a fhir resource for add Blood Pressure', async () => {
+    const resource = service.createResourceData(bloodPressure);
+    if (resource.component) {
+      const diastolicBP = resource.component[0].valueQuantity?.value;
+      const systolicBP = resource.component[1].valueQuantity?.value;
+      expect(diastolicBP).toEqual(bloodPressure.diastolic)
+      expect(systolicBP).toEqual(bloodPressure.systolic)
+    }
+  })
```
semicolin, line 129: Be careful when putting `expect()` inside an `if`. This test will pass if `createResourceData` returns null.

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.spec.ts
```diff
@@ -3,13 +3,18 @@ import { FhirDataService } from './fhir-data.service';
...
 
 describe('FhirDataService', () => {
   let service: FhirDataService;
 
   beforeEach(() => {
     sessionStorage.clear();
-    TestBed.configureTestingModule({});
+    TestBed.configureTestingModule({
+      providers: [{ provide: ReportBPComponent, useValue: bloodPressure }]
+    });
```
semicolin, line 17: Why is this providing an object constant for a component token? Does this actually work? Is anything in this test injecting this?

semicolin (changes requested):

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.spec.ts
```diff
@@ -137,4 +142,38 @@ describe('FhirDataService', () => {
...
+      });
+      if (!service.client) {
+        throw new Error('Client should be initialized');
+      }
+    });
+
+    it('should add patient BP on FHIR server', async () => {
+      const create = spyOn(service.client!, 'create').and.resolveTo(null);
+      const resource: fhirclient.FHIR.Resource = service.createResourceData({ systolic: 1, diastolic: 2 });
+      spyOn(service, 'addPatientData').and.callThrough();
```
semicolin, line 173: This `addPatientData` spy is unnecessary

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.ts
```diff
@@ -94,6 +99,82 @@ export class FhirDataService {
...
     });
   }
+
+  addPatientData(resource: fhirclient.FHIR.Resource) {
+    return this.client?.create(resource);
+  }
+
+  createResourceData(reportBPValue: BloodPressure): fhirclient.FHIR.Resource {
+    return {
+      "resourceType": "Observation",
```
semicolin, line 109: Can we make this method name more specific? How about `createBloodPressureResource()`

semicolin (changes requested):
This also needs a unit test in ReportBPComponent to check that template is wired up correctly so that FhirDataService gets called with correct input values when the button is clicked.

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.spec.ts
semicolin: I checked this and nothing in this test is injecting ReportBPComponent. This can be removed.

semicolin:
When running the unit tests myself, I see some errors in the console from `report-bp.component.spec.ts`. Looks like you missed MatCardModule in the test module imports.

projects/ngx-charts-on-fhir/src/lib/fhir-data/fhir-data.service.spec.ts
```diff
@@ -3,13 +3,18 @@ import { FhirDataService } from './fhir-data.service';
 import FHIR from 'fhirclient';
 import { fhirclient } from 'fhirclient/lib/types';
 import { cold, getTestScheduler } from 'jasmine-marbles';
+import { ReportBPComponent } from 'projects/cardiovascular-patient-app/src/app/report-bp/report-bp.component';
```
semicolin, line 6: please remove this import

semicolin (changes requested):

semicolin (approved):

PR #111, opened 2023-02-28
RT-657-Change-tab

yashpmohod:
## Overview
Use @output decorator and EventEmitter to emit the event.
After the form submit emit the event and pass selectedIndex for the tab is equal to 1.

## How it was tested
Tested Using run cardio-patient-app locally.

## Checklist

### DIFF
--- a/projects/cardiovascular-patient-app/src/app/app.component.html
+++ b/projects/cardiovascular-patient-app/src/app/app.component.html
@@ -5,7 +5,7 @@
   (selectedIndexChange)="selectedIndex = $event">
   <mat-tab label="Report BP">
     <last-report-bp></last-report-bp>
-    <report-bp (resourceCreated)="selectedIndex = $event"></report-bp>
+    <report-bp (resourceCreated)="selectedIndex = 1"></report-bp>
   </mat-tab>
   <mat-tab label="See prior BPs">
     <div class="layout-main" id="layout-main">
--- a/projects/cardiovascular-patient-app/src/app/app.component.html
+++ b/projects/cardiovascular-patient-app/src/app/app.component.html
@@ -1,10 +1,11 @@
 <mat-toolbar>
   <span>{{ appTitle }}</span>
 </mat-toolbar>
-<mat-tab-group mat-stretch-tabs="false" mat-align-tabs="start">
+<mat-tab-group mat-stretch-tabs="false" mat-align-tabs="start" [selectedIndex]="selectedIndex"
+  (selectedIndexChange)="getSelectedIndex($event)">
   <mat-tab label="Report BP">
     <last-report-bp></last-report-bp>
-    <report-bp></report-bp>
+    <report-bp (selectedIndex)="getSelectedIndex($event)"></report-bp>
   </mat-tab>
   <mat-tab label="See prior BPs">
     <div class="layout-main" id="layout-main">
--- a/projects/cardiovascular-patient-app/src/app/app.component.html
+++ b/projects/cardiovascular-patient-app/src/app/app.component.html
@@ -2,10 +2,10 @@
   <span>{{ appTitle }}</span>
 </mat-toolbar>
 <mat-tab-group mat-stretch-tabs="false" mat-align-tabs="start" [selectedIndex]="selectedIndex"
-  (selectedIndexChange)="getSelectedIndex($event)">
+  (selectedIndexChange)="selectedIndex = $event">
   <mat-tab label="Report BP">
     <last-report-bp></last-report-bp>
-    <report-bp (selectedIndex)="getSelectedIndex($event)"></report-bp>
+    <report-bp (resourceCreated)="selectedIndex = $event"></report-bp>
   </mat-tab>
   <mat-tab label="See prior BPs">
     <div class="layout-main" id="layout-main">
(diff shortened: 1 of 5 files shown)


projects/cardiovascular-patient-app/src/app/report-bp/report-bp.component.ts
```diff
@@ -19,6 +19,8 @@ export class ReportBPComponent implements OnInit {
   submitted = false;
   min = 11;
   max = 250;
+  
+  @Output() selectedIndex = new EventEmitter();
```
semicolin, line 23: Please rename this output to something like "bpSubmitted" or "resourceCreated". This component does not have any concept of selected index.

projects/cardiovascular-patient-app/src/app/app.component.ts
```diff
@@ -21,4 +22,8 @@ export class AppComponent implements OnInit {
       layers.forEach((layer) => this.layerManager.select(layer.id));
     });
   }
+
+  getSelectedIndex(index: any) {
+    this.selectedIndex = index
+  }
```
semicolin, line 28: please rename this setter to avoid confusion, or just set the value directly in the template

projects/cardiovascular-patient-app/src/app/app.component.spec.ts
```diff
@@ -1,22 +1,74 @@
...
+    const tabHarness = await tabGroup.getTabs();
+    const systolicInputHarness = await loader.getHarness(MatInputHarness.with({ selector: "[id='systolic']" }));
+    await systolicInputHarness.setValue('110');
+    const diastolicInputHarness = await loader.getHarness(MatInputHarness.with({ selector: "[id='diastolic']" }));
+    await diastolicInputHarness.setValue('70');
+    const submitButtonHarness = await loader.getHarness(MatButtonHarness.with({ selector: "[id='submit']" }));
+    await submitButtonHarness.click();
+    component.getSelectedIndex(1)
+    const selectedTab = await tabHarness[1].isSelected();
+    expect(selectedTab).toBe(true);
```
semicolin, line 50: did you check if this test fails when you break code (e.g, by not emitting the event)?

semicolin (changes requested):
I will work on an example of AppComponent test that mocks the child components. See other comments below...

projects/cardiovascular-patient-app/src/app/app.component.spec.ts
```diff
@@ -1,22 +1,87 @@
...
+@Component({ selector: 'fhir-chart-summary' })
+class MockFhirChartSummaryComponent { }
+
+@Component({ selector: 'last-report-bp' })
+class MockLastReportBPComponent { }
+
+@Component({ selector: 'report-bp' })
+class MockReportBPComponent {
+  @Output() resourceCreated = new EventEmitter<void>();
+}
```
semicolin, line 33: This EventEmitter should have the same type as the real component

projects/cardiovascular-patient-app/src/app/report-bp/report-bp.component.ts
```diff
@@ -19,6 +19,8 @@ export class ReportBPComponent implements OnInit {
   submitted = false;
   min = 11;
   max = 250;
+  
+  @Output() resourceCreated = new EventEmitter();
```
semicolin, line 23: Please add the type param for this EventEmitter, otherwise it defaults to `any`.
If it is emitting the number of resources that were created, it should probably be `EventEmitter<number>`

semicolin (changes requested):

projects/cardiovascular-patient-app/src/app/app.component.html
```diff
@@ -1,10 +1,11 @@
 <mat-toolbar>
   <span>{{ appTitle }}</span>
 </mat-toolbar>
-<mat-tab-group mat-stretch-tabs="false" mat-align-tabs="start">
+<mat-tab-group mat-stretch-tabs="false" mat-align-tabs="start" [selectedIndex]="selectedIndex"
+  (selectedIndexChange)="selectedIndex = $event">
   <mat-tab label="Report BP">
     <last-report-bp></last-report-bp>
-    <report-bp></report-bp>
+    <report-bp (resourceCreated)="selectedIndex = $event"></report-bp>
```
semicolin, line 8: This should not be using `$event` as a tab index. The child component that emits this event does not know which tab should be selected. It doesn't know anything about the tabs.

semicolin (changes requested):

projects/cardiovascular-patient-app/src/app/report-bp/report-bp.component.ts
yashpmohod: ok colin

yashpmohod: Okay, @semicolin I will add the type params for the EventEmitter.

projects/cardiovascular-patient-app/src/app/app.component.html
yashpmohod: @semicolin child component that emits this event with value 1 means the second tab is selected because for every tab there is an index. and we are emitting.

semicolin: ReportBPComponent should not be telling the AppComponent which tab to select. Keeping track of which tab needs to be selected is AppComponent's responsibility. What if we add an instance of report-bp component that is not inside a tab? Then its output would make no sense. What if we add another tab or change their order? Then we would need to update the report-bp component and make its output emit a different number, even though we did not change anything about that component.

semicolin (approved):
Looks good

PR #112, opened 2023-02-28
RT-662-In-the-cardio-patient-app-CSS: Added css for mobile responsive 

amolht:
## Overview
-Added CSS for make responsive mobile view in fhir-chart and fhir-chart-summary.

## How it was tested
Tested it by launching cardio-patient app locally in diffrent mobile view

## Checklist

csschange
mobileview


projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.css
```diff
@@ -11,3 +11,15 @@
...
+  display: flex;
+}
+
+.flex-wrap {
+  flex-wrap: wrap;
+}
+
+.gap-8 {
+  gap: 8px;
+}
```
semicolin, line 25: This is an anti-pattern. Generally speaking, CSS classes should be semantic. They should represent a design element (or one aspect/state of an element). "summary-card-table" is a good example of a semantic CSS class.

"Utility" classes that set a single property should be kept to a minimum because they become difficult to maintain. We would need to change both the CSS and the HTML every time we want to change one of these properties.

semicolin (changes requested):
There is a gap between your CSS breakpoints where the summary card is off screen because it wrapped to the next line but the chart container is still full height.
image

PR #118, opened 2023-03-08
RT-662:In the cardio-patient app, the chart should be optimized for mobile devices and should be easy to read on a small screen.

amolht:
## Overview
Added a css grid layout for app mobile responsive 

## How it was tested
 Tested it by launching the cardio-patient app locally in different mobile views. 
 Also tested showcase-web locally.

## Checklist

### DIFF
--- a/projects/cardiovascular-patient-app/src/styles.scss
+++ b/projects/cardiovascular-patient-app/src/styles.scss
@@ -59,3 +59,12 @@ body {
 .report-bp-card .mat-mdc-card {
   box-shadow: none;
 }
+
+.bp-datepicker {
+  min-width: 300px;
+}
+
+.bp-datepicker .mat-mdc-form-field-infix {
+  max-width: 180px;
+  width: 100%;
+}
--- a/projects/cardiovascular-patient-app/src/styles.scss
+++ b/projects/cardiovascular-patient-app/src/styles.scss
@@ -35,7 +35,7 @@ body {
   --mat-mdc-snack-bar-button-color: #e8e9ec;
 }
 
-// This will make table fixed with so if text increases it will wrap onto next line and will not add horizotal scrollbar
+// This will make table fixed width so if text increases in column it will wrap onto next line and will not add horizotal scrollbar to the table
 
 #summary-card-table {
   .mat-mdc-table {
@@ -47,7 +47,7 @@ body {
   }
 }
 
-// This will minus the header height from the content to remove unnecessary scrollbar
+// This will minus the header height from the content height to remove unnecessary vertical scrollbar from body
 
 .bp-tab-group {
   height: calc(100% - 60px);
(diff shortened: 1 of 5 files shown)


amolht:
375-Iphone-SE

amolht:
iPhone 12 Pro

amolht:
iPhone XR

amolht:
Samsung Galaxy S8 Plus

projects/cardiovascular-patient-app/src/styles.scss
```diff
@@ -33,4 +33,38 @@ body {
...
+  box-shadow: none;
+}
+
+.bp-datepicker {
+  min-width: 300px;
+}
+
+.bp-datepicker .mat-mdc-form-field-infix {
+  max-width: 180px;
+  width: 100%;
```
semicolin, line 69: why are you using global styles for these? can they be added to the component stylesheet instead?

projects/ngx-charts-on-fhir/src/lib/fhir-chart/range-selector/range-selector.component.css
```diff
@@ -1,13 +1,18 @@
...
+
+.range-selector mat-button-toggle {
+  flex:1;
 }
 
-.range-selector mat-form-field {
-  width: 160px;
+.d-flex {
+  display: flex;
 }
```
semicolin, line 18: can `display: flex` be added to the `.bp-datepicker` class instead of creating a separate class just for this?

semicolin (changes requested):
Please add more detail to "How this was tested". Did you test the other apps? or only patient app? Also see comments below...

projects/cardiovascular-patient-app/src/styles.scss
amolht: @semicolin in this CSS some classes generate dynamically at run time, if we use these classes for CSS in the component level we are required to use::ng-deep  for applying CSS for dynamic CSS, but it is deprecated, without::ng-deep the CSS is not applied for dynamic class at the component level. so we used it without the ng-deep in global CSS.

semicolin (changes requested):
If you remove the unnecessary mat-cards and set the height of `.layout-main` based on the viewport instead of parent container, then you can set `<fhir-chart height=100%>` and you don't need to mess with the `.mat-mdc-tab-body-wrapper`.

`grid-template-rows: auto max-content;` can be used to neatly size the vertical layout.

See https://github.com/elimuinformatics/charts-on-fhir/commit/e7e7bd81be5709739d3de22d9065cc68e76b9aa3

That branch still has issues with the calculated chart height being incorrect when range-selector wraps. You may want to just remove the range-selector wrapping for now because we are going to revisit that soon anyway.

PR #124, opened 2023-03-10
RT-676-add-timespan:Add timespan to Current/Previous column headings

yashpmohod:
## Overview
Add a timespan column in the summary table.
Updated test cases according to changes.
image

## How it was tested
Run cardio-patient-app,cardio-health, and showcase app locally.
Test the test cases for showcase and cardio-health,cardio-patient.

## Checklist


semicolin (changes requested):
Please read the requirements again. We want to change the column headings, not add a new row.

projects/ngx-charts-on-fhir/src/lib/fhir-chart-summary/summary.service.ts
```diff
@@ -21,10 +21,12 @@ export class ScatterDataPointSummaryService implements SummaryService {
   summarize(layer: DataLayer, range: NumberRange): Record<string, string>[] {
     const current = this.stats.getFormattedStatistics(layer, range);
     const prev = this.stats.getFormattedStatistics(layer, previous(range));
+    let timeSpan = current['Days Reported'].split('/');
+    timeSpan = timeSpan[1].split(')')
```
semicolin, line 25: This is rather brittle and it's more complex that just computing the number of days again here. It's generally not a good practice to parse values from a formatted string unless there's no other way to get that information.

The code to compute number of days from the range is pretty simple:
```
const days = Math.ceil((range.max - range.min) / MILLISECONDS_PER_DAY)
```

If you don't like duplicating it, you could factor it out into a separate function.

semicolin (changes requested):

projects/ngx-charts-on-fhir/src/lib/fhir-chart-summary/summary.service.ts
yashpmohod: okay @semicolin

amolht (approved):

PR #134, opened 2023-03-14
Rt 684 expression changed after it has been checked error

amolht:
## Overview
- Fixed ExpressionChangedAfterItHasBeenCheckedError.

## How it was tested
-Tested showcase on mock-fhir server.
-Tested showcase, cardio-health, cardio-patient on locally.

## Checklist


semicolin (changes requested):
The range selector is now shown when the chart has no data, which we do not want. And the underlying issue is not fixed because if I click on the buttons and then add data to the chart, I still get an ExpressionChangedAfterItHasBeenCheckedError

PR #140, opened 2023-03-20
RT:678_reference range tooltip

yashpmohod:
## Overview
Add a custom tooltip plugin for showing the reference range.
image
image

## How it was tested
Tested using run showcase,cardio-patient, and cardio-health app locally.

## Checklist

### DIFF
--- a/projects/ngx-charts-on-fhir/src/lib/fhir-chart-summary/statistics.service.spec.ts
+++ b/projects/ngx-charts-on-fhir/src/lib/fhir-chart-summary/statistics.service.spec.ts
@@ -1,3 +1,4 @@
+import { findReferenceRangeForDataset } from '@elimuinformatics/ngx-charts-on-fhir';
 import { ScatterDataPoint } from 'chart.js';
 import { DeepPartial } from 'chart.js/types/utils';
 import { AnnotationOptions } from 'chartjs-plugin-annotation';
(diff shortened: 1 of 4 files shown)


yashpmohod:
@semicolin can you please help me to test a buildConfiguration function 
I have written the following but it giving an error to me can you suggest how to test this 

 describe('buildConfiguration', () => {
    let datasets: Dataset[];
    let scales: ChartScales;
    let annotations: ChartAnnotations;
    let layers: ManagedDataLayer[];

    let configService: FhirChartConfigurationService;


```
    beforeEach(() => {
      layers = [
        {
          name: 'a',
          id: 'a',
          enabled: true,
          datasets: [
            {
              data: [
                { x: 10, y: 100 },
                { x: 20, y: 200 },
              ],
            },
          ],
          scale: { id: 'a' },
        },
      ];
      const layerManager: any = { selectedLayers$: of(layers) };
      configService = new FhirChartConfigurationService(layerManager, timeScaleOptions, ngZone);
      configService.chart = jasmine.createSpyObj<Chart>('Chart', ['zoomScale']);
    });

    it('should return the correct tooltip footer text when reference range is defined', () => {
      const config = configService.buildConfiguration(datasets, scales, annotations);

      const tooltipItem = {
        datasetIndex: 0,
        index: 1,
        x: 1642142400000, // January 14, 2022
        y: 25,
      };
      const tooltipContext = [
        {
          datasetIndex: 0,
          index: 1,
          x: 1642142400000,
          y: 25,
          label: 'Dataset 1',
        },
      ];

      const footerText = config?.options?.plugins?.tooltip?.callbacks?.beforeFooter(tooltipContext);

      expect(footerText).toEqual('Reference Range 10 - 20');
    });

    it('should return an empty string when reference range is not defined', () => {
      const config = configService.buildConfiguration(datasets, scales, annotations);

      const tooltipItem = {
        datasetIndex: 1,
        index: 2,
        x: 1642738800000, // January 21, 2022
        y: 25,
      };
      const tooltipContext = [
        {
          datasetIndex: 1,
          index: 2,
          x: 1642738800000,
          y: 25,
          label: 'Dataset 2',
        },
      ];

      const footerText = config.options?.plugins?.tooltip?.callbacks?.beforeFooter?.(tooltipContext) || '';

      expect(footerText).toEqual('');
    });
  });
```

semicolin:
Try something like this:

```ts
    it('should return the correct tooltip footer text when reference range is defined', () => {
      const dataset: Dataset = {
        label: 'Test',
        yAxisID: 'scale',
        data: [
          { x: 10, y: 100 },
          { x: 20, y: 200 },
        ],
      };
      const scale = { id: 'scale', type: 'linear' } as const;
      const annotations: ChartAnnotations = [
        {
          label: { content: 'Test Reference Range' },
          yScaleID: 'scale',
          yMax: 20,
          yMin: 10,
        },
      ];
      const layerManager: any = { selectedLayers$: EMPTY }
      const configService = new FhirChartConfigurationService(layerManager, timeScaleOptions, ngZone);
      const config = configService.buildConfiguration([dataset], { scale }, annotations);
      const tooltipItems = [{ dataset }] as TooltipItem<TimelineChartType>[];
      const beforeFooter = config.options!.plugins!.tooltip!.callbacks!.beforeFooter!.bind({} as any);
      const footerText = beforeFooter(tooltipItems);
      expect(footerText).toEqual('Reference Range 10 - 20');
    });
```

Notes:
- You don't need to emit anything from layerManager if you're calling `buildConfiguration` directly.
- I'm not sure why the `beforeFooter` callback has a `this` type. Callbacks typically use arrow functions which do not have a `this`. I used `bind` to get around the type error.
- Your test cases were passing the wrong type of object into the callback. It should be an array of [TooltipItem](https://www.chartjs.org/docs/3.9.1/api/interfaces/TooltipItem.html). Our implementation only uses the `dataset` property.
- I used `!` non-null assertions instead of `?` to get around the "possibly undefined" type-check on function call. I usually avoid doing this, but I think it's ok here in unit test as long we make sure the test will fail if the function is undefined/null.
- I would take out the ` || ''` that you used in your other test because that will make the test pass if the callback or its return value is undefined.

PR #145, opened 2023-03-22
Fix error: Cannot read properties of undefined

semicolin:
## Overview

- The call to `zoomScale()` in chart setter was unnecessary. Setting the scale min/max is sufficient to set the initial zoom range. The chart will automatically zoom to those bounds on the first update.

## How it was tested

- Ran patient app with Logica Open environment patient id: 31059
- Reloaded the page multiple times, switched to chart tab before/during/after loading data and checked initial zoom range.

## Checklist


yashpmohod (approved):
looks good

PR #148, opened 2023-03-23
RT-691-Legend-sort-order

yashpmohod:
## Overview
Added a custom pipe to sort the Blood Pressure legend in the order displayed in the chart.
Create a sort-by.pipe.ts file for writing logic for the custom pipe.
image
image

## How it was tested
Tested Using run all charts-on-fhir apps locally.

## Checklist

- [ ] The title contains a short meaningful summary

### DIFF
--- a/projects/ngx-charts-on-fhir/src/lib/fhir-chart-legend/fhir-chart-legend.module.ts
+++ b/projects/ngx-charts-on-fhir/src/lib/fhir-chart-legend/fhir-chart-legend.module.ts
@@ -4,9 +4,10 @@ import { FhirChartLegendComponent } from './fhir-chart-legend.component';
 import { MatIconModule } from '@angular/material/icon';
 import { OverlayModule } from '@angular/cdk/overlay';
 import { MatButtonModule } from '@angular/material/button';
+import { SortByPipe } from './sort-by.pipe';
 
 @NgModule({
-  declarations: [FhirChartLegendComponent],
+  declarations: [FhirChartLegendComponent,SortByPipe],
   imports: [CommonModule, MatIconModule, MatButtonModule, OverlayModule],
   exports: [FhirChartLegendComponent],
 })
--- a/projects/ngx-charts-on-fhir/src/lib/fhir-chart-legend/sort-by.pipe.ts
+++ b/projects/ngx-charts-on-fhir/src/lib/fhir-chart-legend/sort-by.pipe.ts
@@ -1,0 +1,16 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { orderBy, sortBy } from 'lodash';
+
+@Pipe({ name: 'sortBy' })
+export class SortByPipe implements PipeTransform {
+
+    transform(value: any[], order: any = '', column: string = ''): any[] {
+        if (!value || order === '' || !order) { return value; } // no array
+        if (!column || column === '') {
+            if (order === 'asc') { return value.sort() }
+            else { return value.sort().reverse(); }
+        } // sort 1d array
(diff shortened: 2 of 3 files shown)


semicolin (commented):
I like the idea of using a pipe. Maybe we can use it in dataset-list component too instead of the current solution. However, I don't think sorting by label will work for all cases. It probably works for category axis, but for linear axis its just a coincidence. We probably need a more specialized sort pipe that looks at the data. I wrote a function in StatisticsService for sorting by Y value, which is the order that they appear on the chart (at least for linear axis). That function could by improved to work with y-values that are strings too, and then we could use it in both places so we have consistent sorting everywhere.

yashpmohod:
> I like the idea of using a pipe. Maybe we can use it in dataset-list component too instead of the current solution. However, I don't think sorting by label will work for all cases. It probably works for category axis, but for linear axis its just a coincidence. We probably need a more specialized sort pipe that looks at the data. I wrote a function in StatisticsService for sorting by Y value, which is the order that they appear on the chart (at least for linear axis). That function could by improved to work with y-values that are strings too, and then we could use it in both places so we have consistent sorting everywhere.

If I use pipe in the dataset-list component instead of `fhir-chart-legend.component` file it gives an error but if we use in `fhir-chart-legend.component` its works properly with the `StatisticsService` function it sorts by Y value not using a label.

projects/ngx-charts-on-fhir/src/lib/fhir-chart-legend/sort-by.pipe.ts
```diff
@@ -0,0 +1,13 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { Dataset } from '../data-layer/data-layer';
+import { byMostRecentValue } from '../fhir-chart-summary/statistics.service';
+
+@Pipe({ name: 'sortBy' })
+export class SortByPipe implements PipeTransform {
```
semicolin, line 6: Let's rename this now that behavior is more specific. How about `sortDatasets`?

semicolin (changes requested):
How does it behave for medications? Can you add a test case for a medications layer?

projects/ngx-charts-on-fhir/src/lib/fhir-chart-legend/sort-by.pipe.ts
yashpmohod: Okay @semicolin

projects/ngx-charts-on-fhir/src/lib/fhir-chart-legend/fhir-chart-legend.module.ts
```diff
@@ -4,9 +4,10 @@ import { FhirChartLegendComponent } from './fhir-chart-legend.component';
 import { MatIconModule } from '@angular/material/icon';
 import { OverlayModule } from '@angular/cdk/overlay';
 import { MatButtonModule } from '@angular/material/button';
+import { SortDatasets } from './sort-by.pipe';
```
semicolin, line 7: filename is still using the old pipe name
```suggestion
import { SortDatasets } from './sort-datasets.pipe';
```

semicolin (changes requested):
Sorry, noticed one more thing...

projects/ngx-charts-on-fhir/src/lib/fhir-chart-legend/fhir-chart-legend.module.ts
yashpmohod: Okay, @semicolin I will update this.

PR #167, opened 2023-04-13
RT-723-Range-selector-buttons-should-use-today-as-end-date

amolht:
## Overview

- updated code for range selector buttons to use today as the end date.
- updated test cases.

## How it was tested
-Tested it by launching showcase, Cardiovascular Health, and cardio-patient locally.
-Ran unit test locally.

## Checklist

### DIFF
--- a/libs/ngx-charts-on-fhir/src/lib/range-selector/range-selector.component.spec.ts
+++ b/libs/ngx-charts-on-fhir/src/lib/range-selector/range-selector.component.spec.ts
@@ -13,8 +13,9 @@ import { MatDatepickerModule } from '@angular/material/datepicker';
 import { MatNativeDateModule } from '@angular/material/core';
 import { FormsModule } from '@angular/forms';
 import { FhirChartConfigurationService } from '../fhir-chart/fhir-chart-configuration.service';
+import { subtractMonths } from './range-selector.component';
 
-const max = new Date('2022-03-30T00:00').getTime();
+const max = new Date().getTime();
 const min = new Date('2022-01-06T00:00').getTime();
 
 class MockConfigService {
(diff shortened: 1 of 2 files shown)


libs/ngx-charts-on-fhir/src/lib/range-selector/range-selector.component.spec.ts
```diff
@@ -59,28 +60,28 @@ describe('RangeSelectorComponent', () => {
   it('should calculate proper 1 month ago date from max layer date', async () => {
     let ButtonInputGroup = await loader.getHarness(MatButtonToggleHarness.with({ selector: "[id='1 mo']" }));
     await ButtonInputGroup.check();
-    const expectedMinDate = new Date('2022-02-28T00:00');
+    const expectedMinDate = subtractMonths(component.maxDate!, 1);
```
semicolin, line 63: > 4. Do not use an output of real code as the expected value in an assertion. It might be incorrect.

These test cases are supposed to check whether the code is calculating the moth difference correctly. If you use that exact same function to compute the expected value in the test code, you're not really testing anything. It's like checking whether X = X

libs/ngx-charts-on-fhir/src/lib/range-selector/range-selector.component.ts
```diff
@@ -78,7 +78,7 @@ export class RangeSelectorComponent {
   }
 }
 
-function subtractMonths(oldDate: Date, months: number): Date {
+export function subtractMonths(oldDate: Date, months: number): Date {
```
semicolin, line 81: > 8. Do not modify the real code to add special inputs/outputs just for the unit test. This includes making private fields public or adding getters/setters.

This module should not export a private function just for its unit tests

semicolin (changes requested):
The end date field no longer updates when zooming the chart. It always shows today's date.

libs/ngx-charts-on-fhir/src/lib/range-selector/range-selector.component.spec.ts
```diff
@@ -57,30 +57,34 @@ describe('RangeSelectorComponent', () => {
   });
 
   it('should calculate proper 1 month ago date from max layer date', async () => {
+    jasmine.clock().mockDate(new Date('2023-04-13'));
     let ButtonInputGroup = await loader.getHarness(MatButtonToggleHarness.with({ selector: "[id='1 mo']" }));
     await ButtonInputGroup.check();
-    const expectedMinDate = new Date('2022-02-28T00:00');
+    const expectedMinDate = new Date('2023-03-13');
```
semicolin, line 63: 1. I chose those dates specifically to test an edge case in the month calculation with February having fewer days. Please don't change the dates. mockDate should use `'2022-03-30T00:00'` and expected date should be the same as before.
2. After changing that, the mock `timelineRange$.max` will be the same as the mocked current date. We should also change the mock timelineRange$ so the tests will fail if the buttons are using `timelineRange$.max`
3. I included the `T00:00` because I wanted these dates to be treated as local time. Date constructor treats date-only string as UTC, which made the tests fail in some timezones. Why did this need to be changed to date-only? Does mockDate affect the timezone?

semicolin (changes requested):

libs/ngx-charts-on-fhir/src/lib/range-selector/range-selector.component.spec.ts
```diff
@@ -57,30 +57,34 @@ describe('RangeSelectorComponent', () => {
   });
 
   it('should calculate proper 1 month ago date from max layer date', async () => {
+    jasmine.clock().mockDate(new Date('2023-04-13'));
```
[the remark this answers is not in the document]
amolht: @semicolin  The mock date is set using the jasmine.clock().mockDate() method, it will only affect the current test case and will not affect any other test cases that run afterwards. Additionally, once the current test case is completed, the mock date will be removed, and the system time will return to its original state. 
also tried using `clock().install()/uninstall` but gets two errors
1.`Failed to find element matching one of the following queries:
        (MatButtonToggleHarness with host element matching selector: ".mat-button-toggle" satisfying the constraints: host matches selector "[id='1 y']")`
  2.`Error: Jasmine Clock was unable to install over custom global timer functions. Is the clock already installed?`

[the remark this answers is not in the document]
semicolin: got it. thanks for the explanation

semicolin (approved):

PR #169, opened 2023-04-14
Medication Chart Improvements

semicolin:
## Overview

- Use floating bar chart for medications instead of line chart
- Use annotation plugin for medication labels instead of y-axis labels so the labels will be visible when they overlap the bars
- All annotations must have an `id`. `DataLayerMergeService` now uses the annotation `id` when merging layers. If a `Mapper` generates multiple annotations with the same `id`, only the first one will be included in the merged layer.
- Adds support for medication label annotations to annotation-options component
- annotation-options component now only displays color, yMax, and yMin fields for `box` annotations.
- Cleaned up the annotation-options form labels
- Removed custom `medication` scale type because it is no longer needed
- Added support for `tooltip` property on data points so each `Mapper` can define custom tooltips
- Fixed tooltips for medications chart
- Added `chartsOnFhir` property to `Dataset` type where we can add custom options
- `DataLayerColorService` uses `chartsOnFhir.backgroundStyle` property to decide whether to apply transparency to chart elements (defaults to 'solid')

## How it was tested

- Ran showcase app with mock FHIR server
- Tested medications layer (tooltips, labels, bounds)
- Tested a layer with both Simple and Duration medications of the same type
- Regression tested Blood Pressure layer (tooltips, annotations, bounds)
- Checked annotation options panel for Medications and Blood Pressure
- Added lots of layers and resized browser window to make sure medication bars display clearly and labels are properly aligned

## Checklist


semicolin:
image
image
image
image

semicolin:
image
image

libs/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart-configuration.service.ts
```diff
@@ -205,7 +205,10 @@ const setScaleBounds = produce((layers: ManagedDataLayer[]) => {
 });
 
 function computeBounds(axis: 'x' | 'y', padding: number, datasets: Dataset[], annotations?: ChartAnnotations): Partial<NumberRange> {
-  const values = datasets.flatMap((dataset) => dataset.data.map((point) => point[axis])).filter(isNotNaN);
+  const values = datasets
+    .flatMap((dataset) => dataset.data.map((point) => point[axis]))
+    .flat()
+    .filter(isNotNaN);
```
semicolin, line 211: I added `.flat()` because the x-value for medications can now be a `[number, number]` tuple

amolht (dismissed):

yashpmohod (approved):

amolht (approved):

yashpmohod (approved):

PR #170, opened 2023-04-17
RT-734-Typescript-decorators-error

yashpmohod:
## Overview
Changed extends the path in the `tsconfig.lib.json` and `tsconfig.json` files.
 
 
## How it was tested
Open charts-on-fhir project in VS code. and run app chart-on-fhir apps locally.

## Checklist


libs/ngx-charts-on-fhir/tsconfig.json
```diff
@@ -1,3 +1,4 @@
 {
-  "extends": "./tsconfig.lib.json"
+  "extends": "../../tsconfig.base.json",
+  "include": ["**/*.ts"]
```
semicolin, line 3: does `tsconfig.lib.prod.json` need to extend `./tsconfig.lib.json` now?

semicolin (changes requested):
Something similar probably needs to be done for each app because I'm still seeing the errors in patient app.component.spec.ts. Also see question below...

libs/ngx-charts-on-fhir/tsconfig.json
yashpmohod: Yes `tsconfig.lib.prod.json` needs to extend `./tsconfig.lib.json` now and also `tsconfig.spec.json` needs to extend `tsconfig.json`.

apps/cardiovascular-patient-app/tsconfig.app.json
```diff
@@ -0,0 +1,5 @@
+{
+  "extends": "./tsconfig.json",
+  "exclude": ["src/test.ts", "src/polyfills.ts", "src/**/*.spec.ts"],
+  "include": ["**/*.ts"]
```
semicolin, line 4: I don't think we need to repeat the "include" here because it's in ./tsconfig.json

semicolin (changes requested):
What about the other apps? Don't they need the same changes as patient-app?

apps/cardiovascular-patient-app/tsconfig.app.json
yashpmohod: Okay, @semicolin I will remove this.

apps/cardiovascular-health/tsconfig.app.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "./tsconfig.json",
+  "exclude": ["src/test.ts", "src/polyfills.ts", "src/**/*.spec.ts"]
```
semicolin, line 3: why are these excluding `src/polyfills.ts`? don't we want polyfills included in the app build?

yashpmohod: Okay, @semicolin  I will remove `src/polyfills.ts` from `tsconfig.app.json` for all `charts-on-fhir apps`.

semicolin (approved):

amolht (approved):

PR #173, opened 2023-04-18
RT-720-Clinic-measurements: added label for non home measurements

amolht:
## Overview

-Added labeled as a clinic for non-home measurements.
-Updated test cases.
## How it was tested
-Tested by launching all charts-on-fhir apps.
-Ran test cases locally.

## Checklist

image
image


amolht:
> It looks like this broke the dataset-matching code in color service. Also, I wonder if those label suffixes could somehow be integrated with the tags?

@colin ok i will check that with your updated code

amolht:
created new PR

PR #174, opened 2023-04-19
RT-720-Clinic-measurements: added clinic label to non home measurment

amolht:
## Overview

-Added labeled as a clinic for non-home measurements.
-Updated test cases.

## How it was tested
-Tested by launching all charts-on-fhir apps.
-Ran test cases locally.

## Checklist

image

### DIFF
--- a/libs/ngx-charts-on-fhir/src/lib/fhir-chart-summary/home-measurement-summary.service.ts
+++ b/libs/ngx-charts-on-fhir/src/lib/fhir-chart-summary/home-measurement-summary.service.ts
@@ -5,6 +5,7 @@ import { ScatterDataPointSummaryService, SummaryService } from './summary.servic
 import { groupBy } from 'lodash-es';
 
 export const HOME_DATASET_LABEL_SUFFIX = ' (Home)';
+export const CLINIC_DATASET_LABEL_SUFFIX = ' (Clinic)';
 
 /** Combines home and office measurements before summarizing */
 @Injectable({
--- a/libs/ngx-charts-on-fhir/src/lib/fhir-mappers/observation/blood-pressure-mapper.service.spec.ts
+++ b/libs/ngx-charts-on-fhir/src/lib/fhir-mappers/observation/blood-pressure-mapper.service.spec.ts
@@ -68,8 +68,8 @@ describe('BloodPressureMapper', () => {
         ],
       };
       expect(mapper.map(observation).annotations).toEqual([
-        jasmine.objectContaining({ label: { content: 'Systolic Blood Pressure Reference Range' } }),
-        jasmine.objectContaining({ label: { content: 'Diastolic Blood Pressure Reference Range' } }),
+        jasmine.objectContaining({ label: { content: 'Systolic Blood Pressure (Clinic) Reference Range' } }),
+        jasmine.objectContaining({ label: { content: 'Diastolic Blood Pressure (Clinic) Reference Range' } }),
       ]);
     });
   });
--- a/libs/ngx-charts-on-fhir/src/lib/fhir-mappers/observation/blood-pressure-mapper.service.ts
+++ b/libs/ngx-charts-on-fhir/src/lib/fhir-mappers/observation/blood-pressure-mapper.service.ts
@@ -6,6 +6,7 @@ import { Mapper } from '../multi-mapper.service';
 import { ANNOTATION_OPTIONS } from '../fhir-mapper-options';
 import { ChartAnnotation } from '../../utils';
 import { DataLayer } from '../../data-layer/data-layer';
+import { getMeasurementSettingSuffix } from './simple-observation-mapper.service';
 
 const bpCodes = ['85354-9'] as const;
 export type BloodPressureObservation = {
@@ -30,7 +31,7 @@ export class BloodPressureMapper implements Mapper<BloodPressureObservation> {
       merge({}, this.annotationOptions, {
         id: 'Systolic Blood Pressure Reference Range',
         display: true,
-        label: { content: 'Systolic Blood Pressure Reference Range' },
+        label: { content: `Systolic Blood Pressure${getMeasurementSettingSuffix(resource)} Reference Range` },
         yScaleID: layer.datasets[0].yAxisID,
         yMax: 140,
         yMin: 90,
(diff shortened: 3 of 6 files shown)


libs/ngx-charts-on-fhir/src/lib/fhir-mappers/observation/blood-pressure-mapper.service.spec.ts
```diff
@@ -68,8 +68,8 @@ describe('BloodPressureMapper', () => {
         ],
       };
       expect(mapper.map(observation).annotations).toEqual([
-        jasmine.objectContaining({ label: { content: 'Systolic Blood Pressure Reference Range' } }),
-        jasmine.objectContaining({ label: { content: 'Diastolic Blood Pressure Reference Range' } }),
+        jasmine.objectContaining({ label: { content: 'Systolic Blood Pressure (Clinic) Reference Range' } }),
+        jasmine.objectContaining({ label: { content: 'Diastolic Blood Pressure (Clinic) Reference Range' } }),
```
semicolin, line 72: why is this expecting (Clinic) to show up in the reference range label? We don't want separate ref ranges for Home/Clinic. There should only be one reference range for Systolic and one for Diastolic.

semicolin (changes requested):
The screenshot is still showing the wrong colors. The color service needs to be updated so it can match datasets correctly when they both have a suffix.

libs/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
```diff
@@ -34,7 +34,7 @@ describe('DataLayerColorService', () => {
       };
       service.chooseColorsFromPalette(layer);
       expect(layer.datasets[0].borderColor).toEqual(palette[0]);
-      expect(layer.datasets[1].borderColor).toEqual(palette[0]);
+      expect(layer.datasets[1].borderColor).toEqual(palette[1]);
```
semicolin, line 37: This test description is `it('should reuse color from matching dataset'`. This change makes it expect the opposite.
Please update the test so it correctly tests the new dataset matching code.

libs/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.spec.ts
```diff
@@ -43,12 +43,21 @@ describe('DataLayerColorService', () => {
         annotations: [{ label: { display: true } }],
       };
       service.chooseColorsFromPalette(layer);
-      expect(layer.datasets[0].borderColor).toEqual('#333333');
+      expect(layer.datasets[0].borderColor).toEqual('#000000');
```
semicolin, line 46: This test description is `it('should brighten palette color by 20% for home measurements'`. Why would this test expect the color to be black if it has been brightened by 20%?
Please update the test so it correctly tests what it is claiming to test.

libs/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.ts
```diff
@@ -89,7 +89,7 @@ export class DataLayerColorService {
       for (let annotation of layer.annotations) {
         const anno = annotation as BoxAnnotationOptions;
         const label = anno.label?.content;
-        if (typeof label === 'string' && label.startsWith(dataset.label)) {
+        if (typeof label === 'string' && dataset.chartsOnFhir?.group && label.startsWith(dataset.chartsOnFhir?.group)) {
           anno.backgroundColor = this.addTransparency(this.getColor(dataset));
         }
```
semicolin, line 94: It's not intuitive that the dataset needs to define a group that matches the annotation label in order for the annotation to be colored correctly. Maybe we should add a `chartsOnFhir.referenceRangeAnnotationId` property that we can use to match the reference range?
(We could create a separate PR for this if you prefer)

libs/ngx-charts-on-fhir/src/lib/fhir-chart-summary/home-measurement-summary.service.ts
```diff
@@ -27,6 +28,8 @@ export class HomeMeasurementSummaryService implements SummaryService {
 export function getOriginalLabel(dataset: Dataset): string | undefined {
   if (dataset.label && dataset.label.endsWith(HOME_DATASET_LABEL_SUFFIX)) {
     return dataset.label.substring(0, dataset.label.length - HOME_DATASET_LABEL_SUFFIX.length);
+  } else if (dataset.label && dataset.label.endsWith(CLINIC_DATASET_LABEL_SUFFIX)) {
+    return dataset.label.substring(0, dataset.label.length - CLINIC_DATASET_LABEL_SUFFIX.length);
   }
```
semicolin, line 33: Can we use `chartsOnFhir.group` for grouping datasets instead of label? 
(We could create a separate PR for this if you prefer)

libs/ngx-charts-on-fhir/src/lib/fhir-mappers/observation/blood-pressure-mapper.service.ts
```diff
@@ -30,15 +29,15 @@ export class BloodPressureMapper implements Mapper<BloodPressureObservation> {
...
         yScaleID: layer.datasets[0].yAxisID,
         yMax: 140,
         yMin: 90,
       }),
       merge({}, this.annotationOptions, {
         id: 'Diastolic Blood Pressure Reference Range',
         display: true,
-        label: { content: 'Diastolic Blood Pressure Reference Range' },
+        label: { content: `Diastolic Blood Pressure Reference Range` },
         yScaleID: layer.datasets[0].yAxisID,
```
semicolin, line 41: why did you change these to backticks when there are no template variables used in these strings?

semicolin (changes requested):

libs/ngx-charts-on-fhir/src/lib/fhir-chart-summary/home-measurement-summary.service.ts
amolht: for chartsOnFhir.referenceRangeAnnotationId property and this issue we will create separate PR

libs/ngx-charts-on-fhir/src/lib/data-layer/data-layer-color.service.ts
```diff
@@ -68,15 +66,20 @@ export class DataLayerColorService {
...
+
+  private datasetColorIndex(color: string) {
+    if (color) {
+      for (let palette of [this.palette, this.lightPalette]) {
+        const index = palette.indexOf(color);
+        if (index >= 0) {
+          return index;
         }
       }
     }
```
semicolin, line 85: You made a subtle change to the behavior of this function when you factored out `datasetColorIndex()`.  `getMatchingDatasetColorIndex()` now returns undefined if the first dataset in the group does not have a color defined (or if it's not in either of the palettes). Previously, it would move on to check the next dataset in the group and would only return null after checking all of them.

semicolin (changes requested):

PR #179, opened 2023-04-24
RT-745-Shorten-Systolic/Diastolic-labels

yashpmohod:
## Overview
In `blood-pressure-mapper.service.ts` change the labels of `annotationOptions`.
Write the `removeBloodPressureLabel` common function in `utils.ts`, For checking and removing `Blood pressure` from the dataset label and annotation label content.

## How it was tested
Tested using run all `charts-on-fhir` apps locally.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,8 @@ FROM node:18.13.0-alpine AS build
 USER node
 WORKDIR /home/node
 COPY --chown=node package*.json ./
-RUN npm ci
+RUN npm ci --ignore-scripts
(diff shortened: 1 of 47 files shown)


yashpmohod:
After removing `Blood Pressure` from the dataset label and annotation label content.
image
image
image

semicolin (changes requested):
We should not add a special case for BP in the list components. We should just change the label for those datasets when mapping them.

semicolin (changes requested):

semicolin:
@yashpmohod This PR is no longer needed. I have another solution for customizing layer names and matching annotations.

PR #190, opened 2023-05-03
RT-725-Asterisks-on-medication-names

amolht:
## Overview

   -Added a group duration and simple medication datasets.
   -Used that in medication summary service for a groupBy and show group name as the label.

## How it was tested
-Tested it by launching showcase, cardio-health locally.
-Ran unit test locally

## Checklist

image


libs/ngx-charts-on-fhir/src/lib/fhir-chart-summary/medication-summary.service.ts
```diff
@@ -12,6 +13,9 @@ export class MedicationSummaryService implements SummaryService {
     return layer.scale.type === 'category';
   }
   summarize(layer: DataLayer<TimelineChartType, MedicationDataPoint[]>): Record<string, string>[] {
+    const groups = groupBy(layer.datasets, (dataset) => dataset.chartsOnFhir?.group ?? dataset.label);
+    const datasets = Object.entries(groups).map(([label, ds]) => ({ ...ds[0], label, data: ds.flatMap((d) => d.data) }));
+    layer = { ...layer, datasets };
     return layer.datasets.map((dataset) => ({
```
semicolin, line 19: reassigning layer is unnecessary. You can just use the datasets var:
```suggestion
    return datasets.map((dataset) => ({
```

libs/ngx-charts-on-fhir/src/lib/fhir-mappers/medication-request/duration-medication-mapper.service.ts
```diff
@@ -207,12 +207,15 @@ export class DurationMedicationMapper implements Mapper<DurationMedication> {
     const layer = this.baseMapper.map(resource) as DataLayer<'bar', MedicationDataPoint[]>;
     layer.datasets = layer.datasets.map((dataset) => ({
       ...dataset,
-      label: dataset.label + ' *',
+      label: dataset.label,
```
semicolin, line 210: I think it will combine simple and duration medications into the same dataset if they have the same label. We don't want that. Did you test this with a patient who has both simple and duration observations of the same medication?

semicolin (changes requested):

libs/ngx-charts-on-fhir/src/lib/fhir-mappers/medication-request/duration-medication-mapper.service.ts
semicolin: I tested this and the duration bars do not show up for medications with mixed simple/duration observations if they have the same dataset label. We need to add something to the label to keep durations in a separate dataset. I suggest: 
```label: dataset.label + ' (Est. Duration)',```

File: libs/ngx-charts-on-fhir/src/lib/fhir-mappers/medication-request/duration-medication-mapper.service.ts [hunk 3]
```diff
@@ -207,12 +207,15 @@ export class DurationMedicationMapper implements Mapper<DurationMedication> {
...
       type: 'bar',
       borderWidth: 1,
       borderSkipped: false,
       barPercentage: 1,
       grouped: false,
+      chartsOnFhir: {
+        group: dataset.label,
+      },
```
semicolin, line 218: I think this is unnecessary because the base (simple) mapper is already setting the group.

semicolin (changes requested):

semicolin (approved):

PR #216, opened 2023-08-30
RT-890:Hide summary range lines

yashpmohod:
## Overview
 *  The chart should not show vertical lines for “1 month ago” and “2 months ago” by default when there is no summary component.

* The summary component should be responsible for adding these annotations.

## How it was tested
 * Tested locally by running showcase, cardio-patient, and cardio-health app
 * Run unit test cases

## Checklist

### DIFF
--- a/libs/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart-configuration.service.ts
+++ b/libs/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart-configuration.service.ts
@@ -29,7 +29,7 @@ export class FhirChartConfigurationService {
     @Inject(TIMEFRAME_ANNOTATION_OPTIONS) private timeframeAnnotationOptions: ChartAnnotation,
     private ngZone: NgZone
   ) {
-    this.setSummaryRange(1);
+    this.setSummaryRange(0);
   }
 
   private timeline: ScaleOptions<'time'> = {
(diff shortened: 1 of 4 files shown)


libs/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart-configuration.service.ts
```diff
@@ -83,6 +83,15 @@ export class FhirChartConfigurationService {
...
   }
 
+  setSummaryRange(months: number) {
+    this.annotationSubject.next([this.buildTimeframeAnnotation('today', 0)]);
+    this.summaryRangeSubject.next({
+      months,
+      max: new Date().getTime(),
+      min: subtractMonths(new Date(), months).getTime(),
+    });
+  }
```
semicolin, line 93: Why is this method still emitting `summaryRangeSubject` if it is not adding the summary range annotations? Seems like the annotations should always be in sync with the summaryRange$ observable.

libs/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart-configuration.service.ts
```diff
@@ -70,7 +70,7 @@ export class FhirChartConfigurationService {
     }
   }
 
-  setSummaryRange(months: number) {
+  showSummaryRange(months: number) {
```
semicolin, line 73: These method names are confusingly similar. Is there a situation where we need to set the summary range but not show it? If not, it might make more sense to have a single `setSummaryRange` method that turns the summary range off if `months` is 0.

semicolin (changes requested):

libs/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart-configuration.service.ts
```diff
@@ -76,11 +76,13 @@ export class FhirChartConfigurationService {
       this.buildTimeframeAnnotation('current', months),
       this.buildTimeframeAnnotation('previous', months * 2),
```
semicolin, line 77: these should only be added if months > 0, otherwise we'll have 3 identical annotations.

semicolin (changes requested):

semicolin (approved):

amolht (approved):

PR #234, opened 2024-02-26
default zoom to 1 year

vdeshkar:
## Overview

- Set the initial zoom range to “1 Year” instead of “All”

## How it was tested

- Tested by launching the charts-on-fhir showcase app, and verified that the graph initially zooms to one year.
- Ran unit test cases on local.

## Checklist

defaultValueTo1y


semicolin (changes requested):
This solution might confuse other developers. When `configService.isAutoZoom` is true, we want the zoom range to automatically fit the data (e.g. when `configService.resetZoom()` is called): 
```
  /** Reset the zoom so it will change automatically to fit the data */
  resetZoom() {
    if (this.chart) {
      this.isZoomRangeLocked = false;
      this.timeline.min = this.timelineDataBounds.min;
      this.timeline.max = this.timelineDataBounds.max;
      if (this.timeline.min != null && this.timeline.max != null) {
        this.chart.zoomScale('x', { min: this.timeline.min, max: this.timeline.max }, 'zoom');
      }
    }
  }
```

This component should not lock the zoom range in response to it being unlocked.

This also breaks the `selectedButton` so the "All" button no longer shows as selected.

PR #244, opened 2024-05-14
Handled the dataLayerService to continue loading data from the other services when one of service the service fails

shubhamparikh927:
## Overview

- Updated the dataLayerService to continue loading data from the other services when one of services throws an error while retrieving data

## How it was tested

- Tested by intentionally failing a service call and launching the charts-on-fhir showcase app locally.

## Checklist

## Screenshot
image

### DIFF
--- a/libs/ngx-charts-on-fhir/src/lib/data-layer/data-layer-manager.service.ts
+++ b/libs/ngx-charts-on-fhir/src/lib/data-layer/data-layer-manager.service.ts
@@ -163,6 +163,7 @@ export class DataLayerManagerService {
   reset() {
     this.cancel$.next();
     this.state = { ...this.state, layers: {}, selected: [] };
+    this.dataRetrievalError$.next(false);
   }
 
   /**
--- a/libs/ngx-charts-on-fhir/src/lib/data-layer/data-layer-manager.service.ts
+++ b/libs/ngx-charts-on-fhir/src/lib/data-layer/data-layer-manager.service.ts
@@ -88,7 +88,7 @@ export class DataLayerManagerService {
     private tagsService: FhirChartTagsService,
     private mergeService: DataLayerMergeService
   ) {}
-  dataRetrievalError = false;
+  dataRetrievalError$ = new BehaviorSubject<boolean>(false);
   private stateSubject = new BehaviorSubject(initialState);
   private get state() {
     return this.stateSubject.value;
@@ -140,7 +140,7 @@ export class DataLayerManagerService {
       ...this.dataLayerServices.map((service) =>
         service.retrieve().pipe(
           catchError((error) => {
-            this.dataRetrievalError = true;
+            this.dataRetrievalError$.next(true);
             console.error(error);
             return EMPTY;
           })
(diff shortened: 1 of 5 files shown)


libs/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.html
```diff
@@ -1,6 +1,7 @@
 <div class="chart-container" [style.width]="width" [style.height]="height">
   <div class="chart-container-resize-fix">
     <ng-container *ngIf="datasets && datasets.length > 0; else empty">
+      <div class="error-msg" *ngIf="layerManager.dataRetrievalError">Some data is missing</div>
```
semicolin, line 4: Adding an element here might mess up the canvas height because it scales to 100% of its container height. Does the x-axis at the bottom of the chart get cut off when the error message is shown?

Also, let's change the error message to "There was a problem retrieving data. The timeline may be incomplete."

libs/ngx-charts-on-fhir/src/lib/data-layer/data-layer-manager.service.ts
```diff
@@ -88,7 +88,7 @@ export class DataLayerManagerService {
     private tagsService: FhirChartTagsService,
     private mergeService: DataLayerMergeService
   ) {}
-
+  dataRetrievalError = false;
```
semicolin, line 91: I don't know if this will work with on-push components. We may need to use a subject so we can notify the component when it changes.

semicolin (changes requested):

libs/ngx-charts-on-fhir/src/lib/fhir-chart/fhir-chart.component.html
shubhamparikh927: Yes, the chart is cut off at the bottom when the message is added. I will update it.

> Adding an element here might mess up the canvas height because it scales to 100% of its container height. Does the x-axis at the bottom of the chart get cut off when the error message is shown?
> 
> Also, let's change the error message to "There was a problem retrieving data. The timeline may be incomplete."

libs/ngx-charts-on-fhir/src/lib/data-layer/data-layer-manager.service.ts
```diff
@@ -136,17 +136,23 @@ export class DataLayerManagerService {
   retrieveAll() {
     this.reset();
     this.loading$.next(true);
-    merge(...this.dataLayerServices.map((service) => service.retrieve()))
+    merge(
+      ...this.dataLayerServices.map((service) =>
+        service.retrieve().pipe(
+          catchError((error) => {
+            this.dataRetrievalError$.next(true);
```
semicolin, line 143: this should also be reset to false in `reset()` method

semicolin (changes requested):

semicolin (approved):

yashpmohod (approved):

vdeshkar (approved):

PR #248, opened 2024-05-31
Updated the observation mapper to use valueQuantity.code

shubhamparikh927:
## Overview
- Updated the observation mapper to use valueQuantity.code
- Updated the unit test case

## How it was tested
- Tested by running showcase app locally

## Checklist

### DIFF
--- a/libs/ngx-charts-on-fhir/src/lib/fhir-mappers/observation/component-observation-mapper.service.spec.ts
+++ b/libs/ngx-charts-on-fhir/src/lib/fhir-mappers/observation/component-observation-mapper.service.spec.ts
@@ -24,7 +24,7 @@ describe('ComponentObservationMapper', () => {
         component: [
           {
             code: { text: 'component' },
-            valueQuantity: { value: 7, unit: 'unit' },
+            valueQuantity: { value: 7, unit: 'unit', code: 'code' },
           },
         ],
       };
@@ -52,11 +52,11 @@ describe('ComponentObservationMapper', () => {
         component: [
           {
             code: { text: 'one' },
-            valueQuantity: { value: 7, unit: 'unit' },
+            valueQuantity: { value: 7, unit: 'unit', code: 'code' },
           },
           {
             code: { text: 'two' },
-            valueQuantity: { value: 8, unit: 'unit' },
+            valueQuantity: { value: 8, unit: 'unit', code: 'code' },
           },
         ],
       };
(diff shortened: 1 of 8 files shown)


semicolin:
This is a potentially high-impact change. I'm going to do some testing on this branch.

semicolin:
I don't think we should merge this PR. It could result in inaccurate information being displayed in the chart. For example, the demo patient "Ocelot Apricot" has multiple hemoglobin observations. Some are measured in `mmol/L` and others are measured in `g/dL`. In this branch, they both show up on the same scale, which is labelled `g/dL`. The measurement of `15 mmol/L` is misrepresented as `15 g/dL`. Those are not equivalent. We would need to do unit conversion in order to display those on the same chart. 

Instead, we should change the observation mappers to use `resource.valueQuantity.code` instead of `resource.valueQuantity.unit` for the scale label. That will eliminate duplicate layers that are caused by cosmetic differences in the human-readable display units without merging layers that really use different units of measure.

We should also continue troubleshooting the bugs that we see when there are two layers with the same name. We should be able to support multiple layers with the same name. Maybe it is actually an issue with the dataset labels being the same? Dataset labels do need to be unique due to chart.js requirement.

libs/ngx-charts-on-fhir/src/lib/fhir-mappers/observation/component-observation-mapper.service.ts
```diff
@@ -81,7 +81,7 @@ export class ComponentObservationMapper implements Mapper<ComponentObservation>
       })),
       scale: merge({}, this.linearScaleOptions, {
         id: layerName,
-        title: { text: [layerName, resource.component[0].valueQuantity.unit] },
+        title: { text: [layerName, resource.component[0].valueQuantity.code] },
```
semicolin, line 84: need to do the same for SimpleObservationMapper

libs/ngx-charts-on-fhir/src/lib/fhir-mappers/observation/component-observation-mapper.service.spec.ts
```diff
@@ -108,7 +108,7 @@ describe('ComponentObservationMapper', () => {
         component: [
           {
             code: { text: 'component' },
-            valueQuantity: { value: 7, unit: 'unit' },
+            valueQuantity: { value: 7, unit: 'unit', code: 'unit' },
```
semicolin, line 111: we should use different values for `unit` and `code` so the test will fail if the mapper is using the wrong property.

semicolin (changes requested):

semicolin (approved):

vdeshkar (approved):

PR #264, opened 2025-01-02
Fix runtime error in medication mapper when data array is empty

vdeshkar:
## Overview

- Updated plugin to ignore datasets with no data

## How it was tested

- Tested by using charts-on-fhir package locally with sapphire.

- Tested by running unit tests .

## Checklist

### DIFF
--- a/libs/ngx-charts-on-fhir/src/lib/fhir-mappers/medication-request/simple-medication-mapper.service.ts
+++ b/libs/ngx-charts-on-fhir/src/lib/fhir-mappers/medication-request/simple-medication-mapper.service.ts
@@ -78,7 +78,9 @@ export class SimpleMedicationMapper implements Mapper<SimpleMedication> {
       afterDatasetsDraw: (chart) => {
         const ctx = chart.ctx;
         chart.data.datasets.forEach((dataset, datasetIndex) => {
-          if (dataset.data.length === 0) return;
+          if (dataset.data.length === 0) {
+            return;
+          }
           const meta = chart.getDatasetMeta(datasetIndex);
           if (meta.yAxisID === 'medications') {
             const label = dataset.label;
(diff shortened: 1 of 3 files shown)


libs/ngx-charts-on-fhir/src/lib/fhir-mappers/medication-request/simple-medication-mapper.service.ts
```diff
@@ -75,6 +78,9 @@ export class SimpleMedicationMapper implements Mapper<SimpleMedication> {
       afterDatasetsDraw: (chart) => {
         const ctx = chart.ctx;
         chart.data.datasets.forEach((dataset, datasetIndex) => {
+          if (dataset.data.length === 0) {
+            return;
+          }
```
semicolin, line 83: Why is this check using `dataset` directly, but the `centerY` calculation needs to call `getDatasetMeta()`?

semicolin (approved):

libs/ngx-charts-on-fhir/src/lib/fhir-mappers/medication-request/simple-medication-mapper.service.ts
vdeshkar: As a meta object has all the data about the chart properties and is used to calculate for centerY, It's better to use a meta object for comparison of data instead of comparing with the dataset,
I have updated and tested it.

semicolin (dismissed):
I updated the PR description to better describe the change to this library, so the release notes will make more sense.

yashpmohod (dismissed):

shubhamparikh927 (commented):
Please update the version to publish the changes.

yashpmohod (approved):

shubhamparikh927 (approved):

PR #269, opened 2025-02-20
Update chartjs-plugin-annotation to version 3.1.0

vdeshkar:
## Overview

-  Updated **chartjs-plugin-annotation** to the latest version **3.1.0** to be compatible with "chart.js": "^4.3.3",

## How it was tested

- Tested by running cardio, cardio-patients, showcase, and documentation app locally with mock data.
- Tested by running unit tests for all apps.

## Checklist

### DIFF
--- a/libs/ngx-charts-on-fhir/src/lib/fhir-mappers/observation/component-observation-mapper.service.ts
+++ b/libs/ngx-charts-on-fhir/src/lib/fhir-mappers/observation/component-observation-mapper.service.ts
@@ -45,9 +45,7 @@ export function isComponentObservation(resource: Observation): resource is Compo
   );
 }
 /** Maps a FHIR Observation resource that has multiple components */
-@Injectable({
-  providedIn: 'root',
-})
+@Injectable()
 export class ComponentObservationMapper implements Mapper<ComponentObservation> {
   constructor(
     @Inject(LINEAR_SCALE_OPTIONS) private readonly linearScaleOptions: ScaleOptions<'linear'>,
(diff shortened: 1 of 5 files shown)


libs/ngx-charts-on-fhir/src/lib/fhir-mappers/observation/component-observation-mapper.service.ts
```diff
@@ -45,7 +45,9 @@ export function isComponentObservation(resource: Observation): resource is Compo
   );
 }
 /** Maps a FHIR Observation resource that has multiple components */
-@Injectable()
+@Injectable({
+  providedIn: 'root',
+})
```
semicolin, line 50: Why was this changed to provide in root? Mappers should not be provided in the root injector. Apps should use `provideChartsOnFhir(withMappers(...))`

libs/ngx-charts-on-fhir/src/lib/patient-browser/patient.service.ts
```diff
@@ -4,7 +4,9 @@ import { EMPTY, filter, map, ReplaySubject, scan, shareReplay } from 'rxjs';
 import { FhirDataService } from '../fhir-data/fhir-data.service';
 import { isDefined } from '../utils';
 
-@Injectable()
+@Injectable({
+  providedIn: 'root',
+})
```
semicolin, line 9: Why was this changed to provided in root?

semicolin (changes requested):

libs/ngx-charts-on-fhir/src/lib/patient-browser/patient.service.ts
vdeshkar: While testing the cardio app with mock data I was getting NullInjectorError for PatientService, So I have added that, but now we can add that to the providers array as we do in the showcase app.

libs/ngx-charts-on-fhir/src/lib/fhir-mappers/observation/component-observation-mapper.service.ts
vdeshkar: Yes, you are right, I made a mistake while running the showcase with mock data. I will remove this.

semicolin (approved):

yashpmohod (approved):

PR #270, opened 2025-02-21
Update to Angular 19

vdeshkar:
## Overview

- Updated to Angular 19.

## How it was tested

- Tested all the applications by running locally with mock data.
- Tested by running unit tests.

## Checklist


apps/cardio-patient/src/app/app.module.ts
```diff
@@ -55,13 +55,16 @@ function initializeFhirClientFactory(service: FhirDataService): () => Promise<vo
     FhirChartTagsLegendComponent,
   ],
   providers: [
-    { provide: APP_INITIALIZER, useFactory: initializeFhirClientFactory, deps: [FhirDataService], multi: true },
+    provideAppInitializer(() => {
+      const initializerFn = initializeFhirClientFactory(inject(FhirDataService));
+      return initializerFn();
```
shubhamparikh927, line 60: Can we can simplify it to
`provideAppInitializer(() => inject(FhirDataService).initialize())`

libs/ngx-charts-on-fhir/src/lib/fhir-data/provide-fhir-initializer.ts
```diff
@@ -1,14 +1,12 @@
...
-    {
-      provide: APP_INITIALIZER,
-      useFactory: initializeFhirFactory(environment),
-      deps: [FhirDataService],
-      multi: true,
-    },
+    provideAppInitializer(() => {
+      const initializerFn = initializeFhirFactory(environment)(inject(FhirDataService));
+      return initializerFn();
+    }),
```
semicolin, line 9: I think this can be simplified to 
```provideAppInitializer(() => inject(FhirDataService).initialize(environment.clientState)),```

File: libs/ngx-charts-on-fhir/src/lib/timeline-range-selector/timeline-range-selector.component.ts [hunk 3]
```diff
@@ -42,6 +41,12 @@ export class TimelineRangeSelectorComponent {
     });
   }
 
+  ngAfterViewChecked(): void {
+    setTimeout(() => {
+      this.changeDetectorRef.detectChanges();
+    }, 0);
+  }
```
semicolin, line 48: This is going to cause a continuous loop of change detection cycles. Why do you think this is needed?

semicolin (changes requested):

libs/ngx-charts-on-fhir/src/lib/timeline-range-selector/timeline-range-selector.component.ts
vdeshkar: You're right. I added it to fix the `ExpressionChangedAfterItHasBeenCheckedError`, but I’ve now resolved the issue by manually setting focus and blur in the test. This way, we avoid the error without causing a continuous loop of change detection.

semicolin (approved):

shubhamparikh927 (approved):

yashpmohod (approved):

PR #291, opened 2026-03-02
Fix sonar security hotspots

yashpmohod:
## Overview
- Updated SonarCloud action version in pull request workflow.
- Modified npm update command in release workflow.

## How it was tested
- Will be tested in the pipeline.

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,7 +26,7 @@ jobs:
 
       # Ensure npm 11.5.1 or later is installed for trusted publishing support
       - name: Update npm
-        run: npm install -g npm@latest
+        run: npm install -g npm@latest --ignore-scripts
 
       - run: npm ci
       - run: npm run build ngx-charts-on-fhir
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,7 @@ permissions:
   contents: read
 
 env:
-  NODE_VERSION: 20.x
+  NODE_VERSION: 22.x
 
 jobs:
   publish:
@@ -23,10 +23,6 @@ jobs:
           node-version: ${{ env.NODE_VERSION }}
           registry-url: https://registry.npmjs.org/
           cache: npm
-
-      # Ensure npm 11.5.1 or later is installed for trusted publishing support
-      - name: Update npm
-        run: npm install -g npm@latest --ignore-scripts
 
       - run: npm ci
       - run: npm run build ngx-charts-on-fhir
--- a/libs/ngx-charts-on-fhir/package.json
+++ b/libs/ngx-charts-on-fhir/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@elimuinformatics/ngx-charts-on-fhir",
-  "version": "9.0.1",
+  "version": "9.0.0",
   "description": "Charts-on-FHIR: A data visualization library for SMART-on-FHIR healthcare applications",
   "license": "Apache-2.0",
   "homepage": "https://elimuinformatics.github.io/charts-on-fhir",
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@elimuinformatics/charts-on-fhir",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
semicolin: this project is not published. we don't need to update the version number

projects/synthea-utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "synthea-utils",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
semicolin: this project is not published. we don't need to update the version number

projects/mock-fhir-server/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mock-fhir-server",
-  "version": "2.0.0",
+  "version": "2.0.1",
```
semicolin: this project is not published. we don't need to update the version number

.github/workflows/release.yml
```diff
@@ -26,7 +26,7 @@ jobs:
 
       # Ensure npm 11.5.1 or later is installed for trusted publishing support
       - name: Update npm
-        run: npm install -g npm@latest
+        run: npm install -g npm@latest --ignore-scripts
```
semicolin: I think this could break the npm update. Can we use a more recent NODE_VERSION so we don't need to manually update npm?

libs/ngx-charts-on-fhir/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@elimuinformatics/ngx-charts-on-fhir",
-  "version": "9.0.0",
+  "version": "9.0.1",
```
semicolin: there are no changes to the library code so we do not need to publish a new version. is this just to test the release workflow?

semicolin (changes requested):

libs/ngx-charts-on-fhir/test-npm-package.sh
```diff
@@ -4,33 +4,23 @@
 
 set -e
 
-# Set versions explicitly for testing
-MIN_ANGULAR_VERSION="20.3.15"
-MIN_ANGULAR_CLI_VERSION="20.3.13"
-MIN_MATERIAL_VERSION="20.2.14"
-echo Using Angular CLI $MIN_ANGULAR_CLI_VERSION
+ANGULAR_CLI_VERSION="${ANGULAR_CLI_VERSION:-latest}"
```
semicolin: we were explicitly setting the Angular version so we could be sure that the library still works on an app that's using our minimum supported version. If we only test on latest, we might accidentally break compatibility with a previous version.

semicolin (changes requested):

libs/ngx-charts-on-fhir/test-npm-package.sh
```diff
@@ -4,33 +4,25 @@
 
 set -e
 
-# Set versions explicitly for testing
-MIN_ANGULAR_VERSION="20.3.15"
 MIN_ANGULAR_CLI_VERSION="20.3.13"
 MIN_MATERIAL_VERSION="20.2.14"
 echo Using Angular CLI $MIN_ANGULAR_CLI_VERSION
+echo Using Angular CLI $ANGULAR_CLI_VERSION
```
semicolin: where is `$ANGULAR_CLI_VERSION` set?

semicolin (dismissed):

semicolin (approved):

shubhamparikh927 (approved):

